PR #93, opened 2015-07-01
Adds syslogging when RAWX is unavailable

conradkleinespel:
Hello everyone at OpenIO :-)

This avoids the `rawx-monitor.py` if RAWX is not up and running. Instead, this makes it log an error to syslog explaining what is going on.

Let me know if this is OK or not.

Best regards,
Conrad

### DIFF
--- a/tools/rawx-monitor.py
+++ b/tools/rawx-monitor.py
@@ -19,6 +19,7 @@
 
 import sys
 import urllib2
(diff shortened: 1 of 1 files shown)


jfsmig:
Thank you, Conrad!

This cannot do damage and is cleaner !

PR #143, opened 2015-08-18
Code reorganized

jfsmig:
The goal is to reduce the number of external dependencies in the C SDK.

"liboiocore" has been created with essential parts from metautils and libgridcluster:
* minimal strings helpers
* configuration parsing
* logging management


jfsmig:
Naming conventions have been applied (oio_${module}_feature) and cause a lot of changes (renamings).

PR #150, opened 2015-08-20
m1v2: now able to replace all the properties on a reference with a new set.

jfsmig:
This is consistent with meta2 and sqlx : set + flush.
Flushing and setting an empty array give the same result as calling a delete.


jfsmig:
Not mentioned in the comment: the transaction management around properties operations have been reviwed. Now is there only one transaction for each call to the backend. This saves extra implicit transactions, the I/O ops induced, the locks, etc.

Issue #212, opened 2015-10-09
meta2: segfault on PROPDEL_FILTERS


jkasarherou:
```
#0  0x0000000000000011 in ?? ()
#1  0x00007fb40d5579d0 in ?? ()
#2  0x00007fb40d5579d0 in ?? ()
#3  0x000000000040b6bd in _input_data_clean (input_data=0x7fb3fc086c40)
    at /home/noone/public_git/github.com/jkasarherou/oio-sds/meta2v2/meta2_filter_context.c:67
#4  0x000000000040b81a in meta2_filter_ctx_clean (ctx=0x7fb3fc09d330)
    at /home/noone/public_git/github.com/jkasarherou/oio-sds/meta2v2/meta2_filter_context.c:121
#5  0x0000000000405a1e in meta2_dispatch_all (reply=0x7fb40d557b50, gdata=0x23d92a0, hdata=0x60f740 <M2V2_PROPDEL_FILTERS>)
    at /home/noone/public_git/github.com/jkasarherou/oio-sds/meta2v2/meta2_gridd_dispatcher.c:78
#6  0x00007fb4115ad71a in _client_call_handler () at /home/noone/public_git/github.com/jkasarherou/oio-sds/server/transport_gridd.c:729
#7  0x00007fb4115adf65 in _client_manage_l4v (client=0x7fb3f0001a00, gba=0x7fb3fc018630)
    at /home/noone/public_git/github.com/jkasarherou/oio-sds/server/transport_gridd.c:820
#8  0x00007fb4115acf2f in transport_gridd_notify_input (clt=0x7fb3f0001a00)
    at /home/noone/public_git/github.com/jkasarherou/oio-sds/server/transport_gridd.c:504
#9  0x00007fb4115ab7c6 in _notify () at /home/noone/public_git/github.com/jkasarherou/oio-sds/server/network_server.c:1439
#10 0x00007fb4115ab8a0 in _client_manage_input () at /home/noone/public_git/github.com/jkasarherou/oio-sds/server/network_server.c:1463
#11 0x00007fb4115aba61 in _client_manage_event (clt=0x7fb3f0001a00, events=1)
    at /home/noone/public_git/github.com/jkasarherou/oio-sds/server/network_server.c:1533
#12 0x00007fb4115aa96c in _work_on_client (srv=0x23b3100, clt=0x7fb3f0001a00)
    at /home/noone/public_git/github.com/jkasarherou/oio-sds/server/network_server.c:1039
[...]

PR #246, opened 2015-10-31
m2v2: nested func made static

jfsmig:
Even if the function didn't use its wrapper's context, it failed when compiled with recent GCC.


jfsmig:
Targets #212

Issue #303, opened 2015-12-16
oio-meta1-server segfault on exit


jkasarherou:
```
#0  0x00007fbbd108f5f5 in g_mutex_lock () from /usr/lib/libglib-2.0.so.0
#1  0x00007fbbd02de40f in member_lock (m=0x7fbbb4003e10) at /home/noone/public_git/github.com/jkasarherou/oio-sds/sqliterepo/election.c:496
#2  0x00007fbbd02df5ab in step_LeaveElection_completion (zrc=-116, d=0x7fbbb4003e10) at /home/noone/public_git/github.com/jkasarherou/oio-sds/sqliterepo/election.c:981
#3  0x00007fbbcfcb4302 in ?? () from /usr/lib/libzookeeper_mt.so.2
#4  0x00007fbbcfcb46bc in ?? () from /usr/lib/libzookeeper_mt.so.2
#5  0x00007fbbcfcbe581 in ?? () from /usr/lib/libzookeeper_mt.so.2
#6  0x00007fbbce0d44a4 in start_thread () from /usr/lib/libpthread.so.0
#7  0x00007fbbce3d313d in clone () from /usr/lib/libc.so.6
```

jfsmig:
A lot of commits involved here, especially ff41e92e77973bb4384bd897c483716d9258e4f8, 9dd57c29ffadfa6bbae4ff701c10a6b3352ed87e, 6ed1b02ce2f2c15bde409b12445e9e55555cbd51, 7c6f02701fa6b410e93f64cf6e3254d6ff63dea0

PR #318, opened 2015-12-18
sqlx: smoother exit sequence for sqliterepo-based

jfsmig:
Avoids calling zookeeper hooks on closed elections.


jfsmig:
Helps #303 but it doesn't fix it.
Hitting closed elections happened from the Zookeeper hooks as well a asynchronous client calls. The latter still require a fix.

PR #319, opened 2015-12-20
server: more checks around idle/inactive connections.

jfsmig:


jfsmig:
A trace in NOTICE level is generated when at least one connection has been closed.

PR #342, opened 2015-12-31
Fix rdir no content size

chrisz:


rawx-apache2/src/rawx_internals.c
```diff
@@ -134,12 +136,19 @@ void send_chunk_event(const char *type, const dav_resource *resource) {
...
+			resource->info->content.mime_type,
+			resource->info->content.chunk_method,
 			resource->info->chunk.id,
 			resource->info->chunk.hash,
 			resource->info->chunk.position,
 			resource->info->chunk.size);
 
+	if (resource->info->content.size)
+		g_string_append_printf(json,
+				",\"content_size\":\"%s\"",
```
jfsmig: Shouldn't the content_size be written as a JSON integer ? (without quotes)

PR #355, opened 2016-01-07
Fix oio-reset.sh for Opensuse

fvennetier:
Opensuse users are advised to add /usr/sbin in their PATH environment variable before running oio-reset.sh, even if the script will work without it.


jfsmig:
The master has been updated so that all the Travis tests pass. Can you please rebase your branch on it and check it passes the tests ?
Thanks a lot!

PR #376, opened 2016-01-21
Fix some memory leaks in C client

fvennetier:


core/sds.c
```diff
@@ -1503,7 +1505,7 @@ _upload_sequential (struct oio_sds_s *sds, struct oio_sds_ul_dst_s *dst,
 		/* feed the upload queue */
 		if (oio_sds_upload_greedy (ul)) {
 			GRID_TRACE("%s (%p) greedy!", __FUNCTION__, ul);
-			guint8 b[8192];
+			guint8 b[8192];  // FIXME: magic number
```
jfsmig: So what ?

PR #412, opened 2016-02-16
account-server: add possibility to use redis-sentinel

fvennetier:


jkasarherou:
Well that's new @sebastienlapierre, what are u doing here?

sebastienlapierre:
i can revert

PR #420, opened 2016-02-18
unlock the scores (warning: groundbreaking)

jkasarherou:


jfsmig:
We won't accept this as is, because there are more cons than pros:

- it might break the LB that requires to sum all the scores (without overflowing an `uint32_t`) in order to perform the weighted random/round-robin. On our platform with more than 15k services of the same type, a limit of 1M would overflow.
- it just worked and didn't required maintenance, and the (absent) purpose doesn't help to justify the change. 
- scores comprised in [0-100] are easy to read for a human

Though, this is interesting and will take part in a larger redesign of the conscience / LB.

PR #471, opened 2016-03-07
rdir: explicit database create

jkasarherou:
add draft cli for rdir interaction


fvennetier:
There is still a bug if the rawx reference is created but there is no rdir service linked to it.

PR #472, opened 2016-03-07
event-agent: new queue backend and multiprocessing

jkasarherou:


jkasarherou:
meant to be release-2.0

PR #501, opened 2016-03-16
blob-indexer: add interval parameter

jkasarherou:


fvennetier:
The only commit of this PR has been included by hand in current master.

jkasarherou:
yeah PR are so mainstream, let's just merge by hand from now on.

PR #523, opened 2016-04-01
rawx: send events to beanstalkd instead of zeromq

fvennetier:
The implementation is dumb and opens a connection for each event.


jfsmig:
To be closed, but it inspires #524  for something more generic. Thanks, @fvennetier !

PR #524, opened 2016-04-02
events: insulates module, adds beanstalkd implementation

jfsmig:
* rawx: modified to benefit from the oio-event pseudo-object
* sqlx*: idem


jfsmig:
This PR allows an admin to configure an event-agent url as `beanstalk://IP:PORT` in `/etc/oio/sds.conf`. `oio-bootstrap.py` has been modified to make this the default:

> event-agent=beanstalk://127.0.0.1:11300

The previous ZeroMQ implementation is still available and recognized, but now the `ipc://` and `tcp://` prefixes are required.

PR #531, opened 2016-04-05
rawx  + beanstalkd: fixes (design, memleak)

jfsmig:
* events / beanstalkd: better check to detect the queue is empty
* events / beanstalkd: memleak fixed, happened once par queue creation
* rawx: 1 queue / request to force one thread manage its own events
* rawx: minor one-shot memleaks fixed


jfsmig:
:up:

PR #547, opened 2016-04-08
server: log access AFTER answering and log response size

### LINKED ISSUE #546: server/gridd: send the access log *after* the reply generation
https://github.com/open-io/oio-sds/blob/master/server/transport_gridd.c#L543

fvennetier:
Closes #546


jfsmig:
Merci!

PR #552, opened 2016-04-12
proxy: when listing, dups the properties in the body

jfsmig:
Destined to replaces the properties in the headers, both are kept for
the transition.


jfsmig:
@fvennetier :small_red_triangle:

PR #557, opened 2016-04-14
SQLite page size, less syscalls (specially futex), less strlen()

jfsmig:
* SQLite page size not taken into account
* a lot of useless syscalls removed for each M2_PREP
* m2v2: removed a bunch of strlen()


sqlx/sqlx_service.c
```diff
@@ -364,6 +365,10 @@ _configure_backend(struct sqlx_service_s *ss)
 	repository_config.sync_solo = ss->sync_mode_solo;
 	repository_config.sync_repli = ss->sync_mode_repli;
 
+	repository_config.page_size = SQLX_DEFAULT_PAGE_SIZE;
+	if (ss->cfg_page_size > 512)
```
fvennetier: 512 is valid, the test should be `>= 512`

PR #575, opened 2016-05-09
core : add properties modification and container deletion in C SDK

Crashdown59:
add get and set properties method in C SDK
add container removal in C SDK


core/oio_sds.h
```diff
@@ -330,6 +337,27 @@ struct oio_error_s* oio_sds_delete (struct oio_sds_s *sds, struct oio_url_s *u);
 struct oio_error_s* oio_sds_has (struct oio_sds_s *sds, struct oio_url_s *url,
 		int *phas);
 
+
+typedef void (*On_Element) (void *ctx, char *key, char *value);
+	
+/* Get properties of a file : v will be allocated */
+struct oio_error_s* oio_sds_get_content_properties (struct oio_sds_s *sds,
+						    struct oio_url_s *url,
+						    On_Element fct, void* ctx);
```
fvennetier: Please don't mix camel case and snake case (we prefer snake case). And indent with 2 tabs after breaking a line.

core/tool_roundtrip.c
```diff
@@ -26,6 +26,9 @@ License along with this library.
 #include "oio_core.h"
 #include "oio_sds.h"
 
+#define STRING1 "toto"
```
jfsmig: Let's manage random strings (cf. `oio_str_randomize`) instead of hand-coded ones.

core/proxy.c
```diff
@@ -397,6 +406,72 @@ oio_proxy_call_container_get_properties (CURL *h, struct oio_url_s *u,
...
+	GError *err = NULL; 
+	for (cmp = 0; NULL != *(values+cmp) && NULL != *(values+cmp+1)
+		     ; cmp+=2 )
+		{
+			struct http_ctx_s i = {
+				.body = _gs_vprintf ("{\"%s\":\"%s\"}",
+					values [cmp], values [cmp+1])
+			};
+			GString *http_url = _curl_content_url (u, "set_properties");
+			err = _proxy_call (h, "POST", http_url->str, &i, NULL);
```
fvennetier: You should set several properties at once instead of doing one request to the proxy for each property.

core/tool_roundtrip.c
```diff
@@ -268,9 +294,13 @@ main(int argc, char **argv)
...
 	err = _roundtrip_common (client, url, path);
 	if (!err)
-		err = _roundtrip_autocontainer (client, url, path);
-
+	err = _roundtrip_autocontainer (client, url, path);
 	int rc = err != NULL;
+	err = oio_sds_delete_container(client, url);
+	if (err) {
+		g_printerr("error remove container %s", oio_error_message(err));
+		return 4;
```
fvennetier: You should clean the error before returning.

core/proxy.c
```diff
@@ -397,6 +406,94 @@ oio_proxy_call_container_get_properties (CURL *h, struct oio_url_s *u,
...
+
+GError *
+oio_proxy_call_content_set_properties (CURL *h, struct oio_url_s *u,
+				       const char* const *values) {
+	int cmp = 0;
+	GError *err = NULL;
+	gchar *tmp = NULL;
+	gchar *result = NULL;
+	GString *val = NULL;
+	for (cmp = 0; NULL != *(values+cmp) && NULL != *(values+cmp+1)
```
jfsmig: * `cmp` is unused out of the loop. We write C99, and you can declare the variable in the `for` statement.
* `cmp` seems to be just an index. A more intuitive name is required here. Why not just `i` ?
* `*(values+cmp+1)` can be even shortened to `values[i+1]`

```
for (int i=0; values[i] && values[i+1] ;i+=2) {
```
instead of
```
for (cmp = 0; NULL != *(values+cmp) && NULL != *(values+cmp+1) ; cmp+=2 ) {
```

core/proxy.c
```diff
@@ -397,6 +406,94 @@ oio_proxy_call_container_get_properties (CURL *h, struct oio_url_s *u,
...
+		     ; cmp+=2 )
+		{
+			if(cmp != 0) {
+				result = g_strdup_printf("%s,\"%s\":\"%s\"",tmp,values [cmp], values [cmp+1]);
```
jfsmig: :dizzy_face:  There is a much better way to do this!

Just fill a `GString*`. We also provide a way to append a "json <key,value> pair" with proper strings encoding. Use here the `oio_str_gstring_append_json_pair` function from `core/oiostr.h`.

An example here: `core/cs.c` at line 164 (https://github.com/open-io/oio-sds/blob/master/core/cs.c#L164)

core/proxy.c
jfsmig: Don't :repeat: yourself, please :)
Building a JSON set of properties (from a char**) is something that will probably be useful somewhere else, so a function will be welcome.

core/proxy.c
```diff
@@ -397,6 +406,94 @@ oio_proxy_call_container_get_properties (CURL *h, struct oio_url_s *u,
...
+	if (props_str && !*props_str && !err)
+		*props_str = out_ctx.body;
+	else
+		g_string_free(out_ctx.body, TRUE);
+	return err;
+}
+
+GError *
+oio_proxy_call_container_set_properties (CURL *h, struct oio_url_s *u,
+				       const char* const *values) {
```
jfsmig: The `{` is not properly indented. For a uniformity purpose, the opening parenthesis of functions must be on a line.

PR #585, opened 2016-05-13
tools : Add file configuration for oio-bootstrap.py

Crashdown59:
some notes : The file must be in yaml or json.
the json file must finish with .json, yaml with .yml
Some options in oio-reset.sh was removed. The remaining one are : -B, -D, -I, -N, -Z
the configuration file must be passed with the -f option in oio-reset.sh .


jfsmig:
It seems good! Whatever it still requires attention :)
* a lot of commented code
* completely drop the `-B` and `-D` options to `oio-reset.sh`: the value can be fetch from the configuration file. Do not hesitate to use (develop) a side CLI tool to get entries from the file.
* competely drop all the options from oio-bootstrap.py

PR #617, opened 2016-06-08
Python api merge

jkasarherou:


jkasarherou:
this changes a lot of stuff regarding the dependencies
so fellow developers you'll will probably want to reinstall a new python virtualenv from scratch to have a clean environment

PR #618, opened 2016-06-10
m2,events: add sequence number, add event logging filter

fvennetier:


jkasarherou:
is this thing legit?

jkasarherou:
this is not needed anymore

Issue #434, opened 2016-02-23
Split score computation from the main conscience config file


racciari:
Split score computation from the main conscience config file into one or multiple seperate files.
It would make the conscience configuration file much more readable and deployment more flexible.

jkasarherou:
+1

fvennetier:
Fixed by #635.

PR #635, opened 2016-07-01
Review storage policy and service pool configuration

### LINKED ISSUE #190: Load-Balancing: a necessary review to reduce its complexity
Currently, the load-balancing has several drawbacks:
* It plays with `service_info_s`, and it is one of the last parts of the code to do so. `service_info_s` is the structure that notoriously prevent us to use logical ID for services instead of network addresses.
* The current complexity trade-off between time and space clearly targets space. This is a bad idea since one service item is rather small.
* Under certain circumstances, the output is not guaranteed. This can happen if the election criterions target too few services in a huge set of not-targetted services with the same type.
* The implementation is rather hard to read, and harder to extend.
* The API is hard to use.
* The current implementation of the locations is to fuzzy to be really efficient with decent compute times.

I propose to reorganize this API to make it easy to use (simple API), with guaranteed output and response time (better internal structures, stronger locations):
* Build pools of services, gathering all the services pollable in the same call. In other words: compatble  types / tags / class
* In a pool, store the services in per-location-buckets: into a bucket, all the services are considered too close to be polled in the same API call.
* Poll services from a pool, with only parameters the number of services targeted and locations constraints.

fvennetier:
Fix #190.


tools/oio-bootstrap.py
```diff
@@ -459,6 +461,33 @@
...
 
+template_conscience_services = """[pools]
+meta2=${M2_REPLICAS},meta2
+rdir=1,rdir
+account=1,account
+rawx2=2,rawx
+rawx3=3,rawx
+rawx6=9,rawx
+rawx17=17,rawx
+rawxevenodd=1,rawx-even;1,rawx-odd
```
jkasarherou: ?

fvennetier:
Also fixes #434.

jfsmig:
I love this! :heart_eyes:

PR #645, opened 2016-07-13
LB pools improvements + reviewed conscience configuration

fvennetier:


core/lb.c
```diff
@@ -377,10 +406,16 @@ _local_slot__poll (struct oio_lb_slot_s *slot, gboolean masked,
...
+		const char *target, gint mask_shift, struct polling_ctx_s *ctx)
 {
-	GRID_TRACE2("%s pool=%s mask=%d target=%s",
-			__FUNCTION__, lb->name, masked, target);
+	oio_location_t mask = lb->location_mask;
+	if (lb->nearby_mode)
+		mask <<= mask_shift;  // Decrease number of differentiating bits
+	else
+		mask >>= mask_shift;  // Increase number of differentiating bits
+
```
jfsmig: Is it intentional to shift the mask ? If the caller pass a mask whose prefix contains some bits to 0, the result will be unexpected.

fvennetier: I did not see a use case with zeroes in the middle. And as the shift is called in a loop, the most significant bits of the mask have already been used in a previous loop, so it is not a problem to set them to zero.

PR #655, opened 2016-07-26
#655 - Tiering: Add possibility to make filters

Crashdown59:
To begin with storage tiering, put a new parameter tiering_settings_file in your bootstrap config file. this parameter must contain the path of the tiering config yaml file.

The filter is implemented with some filters internally.
2 filters are actually implemented : none, who just returns True and python_script who launch a python script and determine if the filter let him do tiering or not.

2 config files are available : 
- /etc/tiering_config.yml -> exemple of what the tiering_config_file must be
- /etc/tiering_creation_date.py -> exemple of python_script filter, test if the actual time is greater than creation_time + <tiering_time> (in seconds, to define in the tiering_settings_file)

To write new python variables, you have 2 parameters at disposition:
obj_infos : the information of the object
filter_args : the arguments available in the args of the tiering config file


jfsmig:
Because we *need* to implement the filters in the fashion of the Amazon S3 lifecycle (with a per-bucket configuration instead of a single per-namespace configuration file), and because of the (huge) estimation of the time necessary to understand, fix and rebase, we prefer to reimplement it another way.

PR #658, opened 2016-08-02
Add new cache implementation in proxy

Crashdown59:
proxy : add new cache implementation in srv-known, srv-down and resolver
proxy : add new optional parameter to indicate cache parameter file


jkasarherou:
OK to close this?

jfsmig:
Not yet, a part of it will be useful very soon.

PR #669, opened 2016-08-12
Jfs properties cleanup

jfsmig:


Crashdown59:
in proxy/m2_actions.c:702

>static GError * _load_content_from_json_object(struct req_args_s *args, 
>                                                                            struct json_object *jbody, GSList **out) {

The norm is more like this, no? :

> static GError * 
>_load_content_from_json_object(struct req_args_s *args, struct json_object *jbody, GSList **out)
>{

#668 have a lot more change of format too, especially in core/ext.c

PR #671, opened 2016-08-17
tests/functional/conscience/test_lb.py : correct test_choose_4_slot

Crashdown59:
The error was because the proxy shuffle all elements of the list after the conscience return the slots.

Now, the test is testing if there are 3 fast slots and 1 slow slots, not testing only the first and last element port

### DIFF
--- a/tests/functional/conscience/test_lb.py
+++ b/tests/functional/conscience/test_lb.py
@@ -84,10 +84,18 @@ class TestLbChoose(BaseLbTest):
         parsed = resp.json()
         self.assertIsInstance(parsed, list)
         self.assertEqual(4, len(parsed))
-        self.assertGreaterEqual(int(parsed[0]["addr"].split(':')[1]), 8000)
-        # the last one should not be 'fast' since there is only 3
-        # and we don't want duplicates (and there is a default fallback)
-        self.assertLess(int(parsed[3]["addr"].split(':')[1]), 8000)
+        fast_number = 0
+        slow_number = 0
+        for element in parsed:
+            if int(element['addr'].split(':')[1]) >= 8000:
+                fast_number += 1
+            else:
+                slow_number += 1
+        # One of the 4 services should not be 'fast' since there is only 3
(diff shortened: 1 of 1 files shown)


tests/functional/conscience/test_lb.py
```diff
@@ -82,12 +82,20 @@ def test_choose_4_slot(self):
                                         'size': 4})
         self.assertEqual(resp.status_code, 200)
         parsed = resp.json()
+        print parsed
```
fvennetier: Please remove this print (or replace with logging).

tests/functional/conscience/test_lb.py
```diff
@@ -82,12 +82,20 @@ def test_choose_4_slot(self):
...
         self.assertEqual(4, len(parsed))
-        self.assertGreaterEqual(int(parsed[0]["addr"].split(':')[1]), 8000)
+        fast_number = 0
+        slow_number = 0
+        for element in parsed:
+            if int(element['addr'].split(':')[1]) >= 8000:
+                fast_number += 1
+            else:
+                slow_number += 1
         # the last one should not be 'fast' since there is only 3
```
fvennetier: As the test has changed, you should modify this comment.

tests/functional/conscience/test_lb.py
```diff
@@ -84,10 +84,19 @@ def test_choose_4_slot(self):
...
+        slow_number = 0
+        for element in parsed:
+            if int(element['addr'].split(':')[1]) >= 8000:
+                fast_number += 1
+            else:
+                slow_number += 1
+        # One of the four elements should not be fast because there is only
+        # 3 fast slots.
+        # The proxy shuffle elements, so we have to verify if one slot in the
+        # list is slow.
```
fvennetier: Still not clear. Please replace with

> One of the 4 services should not be 'fast' since there is only 3
> in the 'fast' slot, and we don't want duplicates, and
> there is a default fallback on any other service of the same type

PR #672, opened 2016-08-18
core: force usage of old fmemopen with glibc >= 2.22

fvennetier:
Binary mode of fmemopen() has been removed in glibc 2.22.
Without binary mode, fmemopen() tries to add '\0' at the end of the
buffer, but if there is no room it fails an never flushes the writes,
leading to unpredictable behavior.


fvennetier:
Superseded by #677 .

Issue #659, opened 2016-08-02
Properties not saved when set upon upload


jfsmig:
Wether tested in Python or in C, when "user" properties are prepared to be saved upon the upload, they are forgotten and are not reported when showing the content. This is broken, and appeared thanks to @jkasarherou recent tests:exclamation: 

Probably it never worked at all, and a good reminder that the tests rules :grey_exclamation:

jfsmig:
It never appeared with the test of the C client, despite the closed status of #558, since @Crashdown59 implemented the set-upon-upload with 2 separate steps.

PR #673, opened 2016-08-18
core: modify set_properties and get_properties to respect new format

Crashdown59:
core: The properties indicate in upload are set in the upload_commit step

### DIFF
--- a/core/sds.c
+++ b/core/sds.c
@@ -1659,6 +1659,7 @@ oio_sds_upload_commit (struct oio_sds_ul_s *ul)
 		.chunk_method = ul->chunk_method,
 		.append = BOOL(ul->dst->append),
 		.update = BOOL(ul->dst->partial),
+		.properties = ul->dst->properties
 	};
 
 	GRID_TRACE("%s (%p) Saving %s", __FUNCTION__, ul, request_body->str);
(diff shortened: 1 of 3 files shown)


Crashdown59:
see #659

core/sds.c
```diff
@@ -2307,10 +2305,16 @@ oio_sds_get_content_properties(struct oio_sds_s *sds, struct oio_url_s *url,
 	err = (struct oio_error_s*) oio_proxy_call_content_get_properties(sds->h, url, &value);
 	if (err)
 		return err;
-
 	json_object *json = json_tokener_parse(value->str);
 	json_object_object_foreach(json,key,val) {
-		fct(ctx, key, json_object_get_string(val));
+		(void) key;
+		const gchar *json_string = json_object_get_string(val);
+		json_object *json_properties = json_tokener_parse(json_string);
```
fvennetier: Is that really necessary to cast the object to a string and parse it again as a json object? You can probably use `json_object_object_foreach` directly on `val`.
It would also be a good idea to check `key`.

core/sds.c
```diff
@@ -2307,10 +2305,16 @@ oio_sds_get_content_properties(struct oio_sds_s *sds, struct oio_url_s *url,
...
+		json_object_object_foreach(json_properties,
+					   key_properties, value_properties)
```
fvennetier: key_properties -> property_key
value_properties -> property_key

PR #676, opened 2016-08-19
all-requirements.txt: replace cliff version dependencies by choosing only 2.0.X versions

Crashdown59:


jkasarherou:
the fix is ok,
but travis build failed because of failure in LB?

fvennetier:
This LB failure has been there for weeks. @Crashdown59 is working on it.

PR #681, opened 2016-08-26
meta2: adding recovery of location with the load balancer

Crashdown59:
correcting problem with chunk move tests


jfsmig:
Superseded with #683

PR #694, opened 2016-09-02
events: convert modification time to seconds, fix syslog prefix (fix #689)

### LINKED ISSUE #689: Blob rebuilder timestamp issue
Using this documentation to deploy a cluster: http://docs.openio.io/install-guide-centos, I have some troubles between the `oio-blob-indexer` and the ```openio volume incident``` command.

Indeed, when I use ```openio volume admin incident 10.0.0.16:7004``` to set an incident on a volume (incident date is then the current timestamp, confirmed by ```openio volume admin show 10.0.0.16:7004```), if then I launch the ```oio-blob-rebuilder --log-level DEBUG OPENIO 10.0.0.16:7004``` nothing happens.

But, if I force the incident date with a timestamp in the future, it works (```openio volume admin incident 10.0.0.16:7004 1471354244```).

Maybe a problem of timezone between the `oio-blob-indexer` and the `openio volume incident`command.

I can provide access to a platform with the problem if needed.

fvennetier:


jkasarherou:
should also check the time format on the rdir side

PR #700, opened 2016-09-14
oio/directory/meta0: correcting rebalance function

Crashdown59:
The bug was caused by the non-check of the length of the random prefixes, who can be zero.

When the prefix set was zero, all prefixes of the service selected were lost, causing a lost of prefixes managed by a meta1


jfsmig:
Fixed in commit d52f907f977802c8e21920849e91be334d0ce804

PR #706, opened 2016-09-21
Various fixes around LB pools

jfsmig:
* core: ceinture
* metautils: bretelles

### DIFF
--- a/core/lb.c
+++ b/core/lb.c
@@ -403,6 +403,10 @@ _local_slot__poll (struct oio_lb_slot_s *slot, oio_location_t mask,
 			__FUNCTION__, slot->name, slot->sum_weight, slot->items->len,
 			mask);
 
+	if (slot->items->len == 0) {
+		GRID_TRACE2("%s slot empty", __FUNCTION__);
+		return FALSE;
+	}
 	if (slot->sum_weight == 0) {
 		GRID_TRACE2("%s no service available", __FUNCTION__);
 		return FALSE;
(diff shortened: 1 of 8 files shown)


core/lb.c
```diff
@@ -929,6 +933,9 @@ void
 oio_lb_world__feed_slot (struct oio_lb_world_s *self, const char *name,
 		const struct oio_lb_item_s *item)
 {
+	if (!item || item->weight <= 0)
+		return;
```
fvennetier: This is dangerous. This will make the service undiscoverable. You don't wan't undiscoverable, you just wan't unselectable. There is a check in `oio_lb_world__feed_slot_unlocked` for this condition, after the item has been integrated in the _world_.

tests/functional/conscience/test_conscience.py
```diff
@@ -1,6 +1,6 @@
 import logging
 
-from tests.utils import BaseTestCase
+from tests.utils import random_str, BaseTestCase
```
fvennetier: This import is not used (breaks the tests).

core/lb.c
jfsmig: Change reverted. Idem for metautils/lib/lb.c. Juste store zeroed services and ignore them.

tests/functional/conscience/test_conscience.py
jfsmig: Removed

PR #709, opened 2016-09-22
cli: add --full argument for full listing

jkasarherou:


oio/cli/storage/container.py
```diff
@@ -197,10 +203,21 @@ def take_action(self, parsed_args):
...
+        if parsed_args.full_listing:
+            l, meta = self.app.client_manager.storage.container_list(
+                account, **kwargs)
+            listing = l
+            while listing:
+                kwargs['marker'] = listing[-1][0]
+                listing, meta = self.app.client_manager.storage.container_list(
+                    account, **kwargs)
+                if listing:
+                    l.extend(listing)
```
fvennetier: Is it possible to yield the results instead of building a mile-long list?

jkasarherou: I'll check

PR #741, opened 2016-10-27
m2: fix prepare data update when not replicated (fix #739)

### LINKED ISSUE #739: Quota not taken into account by non replicated meta2
Steps to reproduce:
```
export OIO_NS=NS
export OIO_ACCOUNT=myaccount
dd if=/dev/zero of=./12M bs=1M count=12
oio-reset.sh -N $OIO_NS -f etc/bootstrap-SINGLE.yml
# Default value set by oio-reset.sh is 50000000
openio cluster show --format shell | grep container_max_size
openio container create toto
for i in `seq 0 5`; do openio object create toto 12M --name 12M-$i; done
openio container show toto --format shell | grep bytes_usage
```
Expected: `Request error: Container's quota reached (50000000 bytes) (HTTP 500) (STATUS 445)`
Actual: no error

The problem does not occur with replicated meta2.

fvennetier:


meta2v2/meta2_backend.c
```diff
@@ -757,6 +760,20 @@ meta2_backend_add_modified_container(struct meta2_backend_s *m2b,
...
 				_container_state (sq3));
+
+	gchar *account = sqlx_admin_get_str(sq3, SQLX_ADMIN_ACCOUNT);
+	gchar *user = sqlx_admin_get_str(sq3, SQLX_ADMIN_USERNAME);
+	struct oio_url_s *url = oio_url_empty();
+	oio_url_set(url, OIOURL_NS, m2b->ns_name);
+	oio_url_set(url, OIOURL_ACCOUNT, account);
+	oio_url_set(url, OIOURL_USER, user);
+
+	_meta2_backend_force_prepare_data(m2b,
```
jfsmig: I suggest avoiding an HEXID compuation with partial information.
Please, wrap the call to  `_meta2_backend_force_prepare_data()` with a guard on `account` and `user`. Both must be non-NULL to ensure the right HEXID is computed.

PR #743, opened 2016-10-31
Fix download ranges

fvennetier:


tests/utils.py
```diff
@@ -37,6 +37,12 @@ def random_id(n):
     return random_str(n, chars=random_chars_id)
 
 
+def random_data(size):
+    """Return `size` bytes of random data as a str object"""
+    with open("/dev/urandom", 'rb') as urandom:
```
jkasarherou: this is not portable

PR #750, opened 2016-11-04
List object properties while listing objects

fvennetier:


core/oio_sds.h
```diff
@@ -384,6 +384,7 @@ struct oio_sds_list_param_s
 
 	char flag_nodeleted : 1;
 	char flag_allversions : 1;
+	char flag_properties : 1;
```
jfsmig: It is a good idea to use "unsigned" variable for flag sets, at least to ease further printing/debug effort.
Could you benefit from the current enhancement effort to also turn these existing "signed char" flags into their "unsigned" equivalent?

core/oio_sds.h
```diff
@@ -382,8 +382,9 @@ struct oio_sds_list_param_s
 	/* 0 means no set */
 	char delimiter;
 
-	char flag_nodeleted : 1;
-	char flag_allversions : 1;
+	unsigned char flag_nodeleted : 1;
+	unsigned char flag_allversions : 1;
+	unsigned char flag_properties : 1;
```
jfsmig: An exposed structure changed, I suggest to alter the API version number.

PR #761, opened 2016-11-18
UDP preferred for DB_USE requests

jfsmig:
Only enabled when explicitely allowed in the `/etc/oio/sds.conf` files, with en entry like this in the namespace's section

    [MyNS]
    udp_allowed = true

The goal is to reduce the overall load with a simpler protocol for dummy tasks.

### DIFF
--- a/sqlx/sqlx_service.c
+++ b/sqlx/sqlx_service.c
@@ -53,6 +53,8 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
 # define SQLX_MAX_TIMER_PER_ROUND 100
 #endif
 
+static volatile gboolean udp_allowed = FALSE;
+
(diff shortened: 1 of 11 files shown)


sqlx/sqlx_service.c
```diff
@@ -543,16 +555,25 @@ sqlx_service_action(void)
 			return _action_report_error(err, "Failed to start the QUEUE thread");
 	}
 
-	/* SERVER/GRIDD main run loop */
+	/* open all the sockets */
 	if (!grid_main_is_running())
 		return;
+	if (udp_allowed)
```
fvennetier: I think it's harmless to enable UDP server by default. This would simplify the migration path: when you know all your servers have been updated, you can enable UDP clients.

PR #764, opened 2016-11-19
Meta1 may now group prefixes in common databases

jfsmig:
meta1 manages prefixes from 0 to 4 digits. This is configured in the `/etc/oio/sds.conf` files.
Elections are per-base, not per prefix, so reducing the number of files has a huge impact of the election load.

* **0** : single `0000.meta1` database file, ideal for minimal demonstrations
* **1** : 16 db files, from 0000.meta1 to F000.meta1, ideal for scalability demonstrations
* **2** : 256 db files, ideal for auto-containers deployments, with less than 8M containers
* **3** : 4096 db files, ideal for big deployments, with less than 256M containers
* **4** : 64ki db files, ideal for huge deployments
* **default**: **4** (for a backward compliance purpose)

Changelog:
* meta1: coalesces prefixes in bases, based on the hexadecimal prefix
* meta1: health check added around coalescence of prefixes, pre-refresh
* oio: `directory bootstrap` manages the variable meta1 prefixes
* tools/oio-bootstrap.py: manages the meta1 variable prefixes
* tools/oio-reset.sh: manages the meta1 variable prefixes
* meta1: code factorized, trimmed, beautified
* etc/bootstrap-*: new presets
* tools/oio-reset.sh: fixes number of directory replicas.
  It was equal to the number of meta1 services, by default.
* meta0: crap code removed, remaining code simplified


oio/cli/clientmanager.py
```diff
@@ -57,6 +58,9 @@ def setup(self):
             self.session = requests.Session()
             self.setup_done = True
             self._admin_mode = self._options.get('admin_mode')
+            # Get the number of meta1 digits from 2 possible sources
```
fvennetier: Where is the second one?

oio/directory/meta0.py
```diff
@@ -37,15 +38,28 @@ class PrefixMapping(object):
     TOTAL_PREFIXES = 65536
 
     def __init__(self, meta0_client, conscience_client, replicas=3,
-                 logger=None, **kwargs):
+                 digits=None, logger=None, **kwargs):
+        """
+        Parameters:
+        - replicas: strictily positive integer
+        - digits: integer between 2 and 4 (default: 4)
```
fvennetier: In the code you actually check range [0, 4].

etc/bootstrap-meta1-0digits.yml
```diff
@@ -0,0 +1 @@
+meta1_digits: 0
```
fvennetier: Need redis for the tests to pass. Also valid for the next 2 files.

oio/cli/clientmanager.py
jfsmig: An abusive comment that remained. The second source imagined was the namespace_info. But that kind of configuration is too dangerous too change on the fly.

meta0v2/meta0_utils.c
```diff
@@ -226,47 +226,47 @@ meta0_utils_unpack_meta1ref(const gchar *s_m1ref,
...
-				return;
-		}
-	}
-
 	g_assert(NULL != bin);
 	g_assert(digits <= 4);
-	return _run(*((const guint16*)bin) & masks[digits], masks[digits]);
+	const guint16 msk_h16 = masks[digits];
+	const guint16 pfx_n16 = *(const guint16*)bin;
+	return _foreach_prefix_run(g_ntohs(pfx_n16) & msk_h16, ~msk_h16, on_prefix, u);
```
fvennetier: The second argument should be `(~msk_h16) + 1` if you want to include the last element (or change the test in the `for` loop).

PR #802, opened 2016-12-14
Fix memory leaks (meta0, metautils)

jfsmig:
The metautils leak specially affects all the sqliterepo-based services.
It happens under heavy loads, when replication RPC generate timeouts on the client side (e.g. "queued for too long").


jfsmig:
It seems a good idea to port both commits to the 3.1.x branch.

PR #808, opened 2016-12-20
Fix some odd mix of memory management, libc vs. glib

jfsmig:
In addition, some more strings handlings have been moved on the stack.

### DIFF
--- a/meta0v2/meta0_client.c
+++ b/meta0v2/meta0_client.c
@@ -28,7 +28,7 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
 #include "meta0_utils.h"
 
 static addr_info_t addr = {{0}};
-static gchar *namespace = NULL;
+static gchar namespace[LIMIT_LENGTH_NSNAME] = {0};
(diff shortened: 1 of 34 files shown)


meta0v2/meta0_client.c
```diff
@@ -28,7 +28,7 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
 #include "meta0_utils.h"
 
 static addr_info_t addr = {{0}};
-static gchar *namespace = NULL;
+static gchar namespace[LIMIT_LENGTH_NSNAME] = {0};
```
fvennetier: Constants like `LIMIT_LENGTH_NSNAME` should be clarified to tell if they include the final `'\0'` character or not. `metautils_message_extract_url` supposes they do not include it.

jfsmig: The usage of the `LIMIT_LENGTH_`  macros is to store the string with its trailing NUL character.
So that the real limit on the length of the information is the macro minus 1.

What do you suggest? A global macro renaming? A fix of the documentation about this?

Issue #806, opened 2016-12-20
event-agent: Second simultaneous read on fileno


jkasarherou:
this is observed in high load when event-agent is configured to notify another tube.
`2016-12-16T16:35:03.731487+09:00 msg-vr-ost1500 OIO,QX-EAST-P,oio-event-agent,0: warning  16620 1818230 log WARNING failed to notify event: Second simultaneous read on fileno 18 detected.  Unless you really know what you're doing, make sure that only one greenthread can read any particular socket.  Consider using a pools.Pool. If you do know what you're doing and want to disable this error, call eventlet.debug.hub_prevent_multiple_readers(False) - MY THREAD=<built-in method switch of GreenThread object at 0x1818230>; THAT THREAD=FdListener('read', 18, <built-in method switch of GreenThread object at 0x18180f0>, <built-in method throw of GreenThread object at 0x18180f0>)`

jkasarherou:
fixed by #811

PR #811, opened 2016-12-21
event-agent fixes

jkasarherou:
should help mitigating #806


jkasarherou:
this PR is reviewing error handling in event-agent

Basically the problem was that retrying actions in filters were unclear
in the filters there should not have `except Exception` that silently captures errors
filters should use a specific error return to retry event
and by default unmanaged exception will bury event

PR #831, opened 2017-01-13
api: fix object_list metadata handling

jkasarherou:


fvennetier:
Hello. You have to remove this line for the tests to pass https://github.com/jkasarherou/oio-sds/blob/0cea50ddf5823f5b089146965b5dd306227eaf83/oio/api/object_storage.py#L20.

Do you think this deserves to be ported to branch 3.1.x? (requires modifications, cannot just cherry-pick)

Issue #859, opened 2017-02-03
one chunk missing in erasurecode  6+3


sebastienlapierre:
Some content missing one chunk

[root@sasopenio01 ~]# openio object show radit_98 myobjects76
+-----------+----------------------------------+
| Field     | Value                            |
+-----------+----------------------------------+
| account   | scalair_account                  |
| container | radit_98                         |
| ctime     | 1486116300                       |
| hash      | A6DCF8423AC55B068C791890527A0B78 |
| id        | 66B76B699D470500F81BE73BDB6A3339 |
| mime-type | application/octet-stream         |
| object    | myobjects76                      |
| policy    | ERASURECODE                      |
| size      | 1024000                          |
| version   | 1486116287657850                 |
+-----------+----------------------------------+
[root@sasopenio01 ~]# openio object locate  radit_98 myobjects76
+-----+-------------------------------------------------------------------------------------------+---------+----------------------------------+
| Pos | Id                                                                                        |    Size | Hash                             |
+-----+-------------------------------------------------------------------------------------------+---------+----------------------------------+
| 0.0 | http://172.16.139.2:6033/5AEEA93908E351112C5179A85FD8537F36959FCFF8941439FDEFD46877D8E787 | 1024000 | A6DCF8423AC55B068C791890527A0B78 |
| 0.1 | http://172.16.139.1:6041/8070665C99435DA93196939CC2E2266B346A4912D961F687DA0AB4A1CA011D41 | 1024000 | A6DCF8423AC55B068C791890527A0B78 |
| 0.2 | http://172.16.139.3:6046/7AED7CF97C23BFDC9CE2D2AB2D21920DBC22CCAB51CD423E4F481479F9E10A54 | 1024000 | A6DCF8423AC55B068C791890527A0B78 |
| 0.3 | http://172.16.139.3:6028/71442BA77B08753C71C2ACAB77363F289FA63E940CF931DAAA6BCAAC0F59190E | 1024000 | A6DCF8423AC55B068C791890527A0B78 |
| 0.4 | http://172.16.139.2:6039/426E86A48E900775243B3E549A3C08188824A6CF46F6B59E099826E3343450B3 | 1024000 | A6DCF8423AC55B068C791890527A0B78 |
| 0.5 | http://172.16.139.1:6025/08296A4897BCDFD640E90D3DBC64C01FC988311346A3292AC642F3C81566CBFF | 1024000 | A6DCF8423AC55B068C791890527A0B78 |
[...]

PR #860, opened 2017-02-06
api: fix missing failed chunk in EC

jkasarherou:


jkasarherou:
fixes #859

oio/api/ec.py
```diff
@@ -648,7 +648,7 @@ def _send(self):
                     self.failed = True
                     msg = str(exc)
                     logger.warn("Failed to write to %s (%s)", self.chunk, msg)
-                    self.chunk['error'] = msg
+                    self.chunk['error'] = 'write: ' % msg
```
fvennetier: There is no format specifier!

PR #864, opened 2017-02-07
proxy: Mention the source error in the access log

jfsmig:
A final token is added on each accesslog linne, stored in the arbitrary part.

```
e={"status":406,"message":"Directory error: no such user"}
```

### DIFF
--- a/metautils/lib/gridd_client.c
+++ b/metautils/lib/gridd_client.c
@@ -343,15 +343,11 @@ _client_manage_reply(struct gridd_client_s *client, MESSAGE reply)
 		return err;
 	}
 
-	/* all other are considered errors
-	 * JFS: if someone want to mangle the message, don't touch it the status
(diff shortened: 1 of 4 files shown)


metautils/lib/gridd_client.c
```diff
@@ -343,11 +343,10 @@ _client_manage_reply(struct gridd_client_s *client, MESSAGE reply)
...
 
-	/* all other are considered errors */
-	if (status != CODE_REDIRECT)
-		err = NEWERROR(status, "Request error: %s", message);
-	else
-		err = NEWERROR(status, "%s", message);
+	/* all other are considered errors
+	 * JFS: if someone want to mangle the message, don't touch it the status
+	 * is equal to CODE_REDIRECT, the message contains the URL as-is. */
+	err = NEWERROR(status, "%s", message);
```
fvennetier: I don't understand that comment.

PR #868, opened 2017-02-08
crawler: add integrity check

jkasarherou:


oio/crawler/integrity.py
```diff
@@ -0,0 +1,334 @@
...
+                      action='store_true', help='verbose output')
+
+    args = parser.parse_args()
+
+    checker = Checker(args.namespace, error_file=args.output)
+    if not os.isatty(sys.stdin.fileno()):
+        source = sys.stdin
+    else:
+        source = cStringIO.StringIO(' '.join(args.target))
+    args = csv.reader(source, delimiter=' ')
```
fvennetier: I suggest to make this delimiter configurable. A single space will cause problems with contents (or containers) with a space in their name.

PR #896, opened 2017-02-27
Big batch of fixes for warnings

jfsmig:
This is mostly the removal of unused code, minor reformatting to ease the task of some IDE.
Please note that:
* metautils now ships its own asn1c code (so we drop the asn1c runtime dependency)
* the proxy receives the fix to a "hopefully impactless" horrible bug.


core/oiostr.h
```diff
@@ -195,6 +195,7 @@ int oio_str_prefixed (const char *s, const char *p, const char *sep);
 
 int oio_str_caseprefixed(const char *str, const char *prefix);
 
+/* pi64 is left untouched is the overall return is FALSE */
```
fvennetier: _is the overall_ -> _if the overall_

PR #912, opened 2017-03-15
Add account listing to the CLI (fix #817)

### LINKED ISSUE #817: Add account listing to the CLI
While using the CLI, it should be great to be able to list all the accounts in a namespace.

Should be something like: `openio account list`

kamel-rahim:


oio/cli/storage/account.py
```diff
@@ -143,3 +143,43 @@ def take_action(self, parsed_args):
...
+        return parser
+
+    def take_action(self, parsed_args):
+        self.log.debug('take_action(%s)', parsed_args)
+
+        data = self.app.client_manager.storage.account_list()
+        accounts = ((e,) for e in data)
+
+        if parsed_args.full_listing:
+            columns = ('Name', 'bytes', 'containers', 'ctime', 'metadata',
```
fvennetier: I would have displayed `metadata` at the end because it's the least important, and `objects` just after `containers`.

PR #921, opened 2017-03-20
proxy: make command lock/unlock check type of the service (fix #822)

### LINKED ISSUE #822: “openio cluster unlock" on non-existant service works
`openio cluster unlock` on non existing service (service name, IP or port) does not return an error.

```
# openio cluster unlock QSDF 192.168.0.251:666
+------+-------------------+----------+
| Type | Service           | Result   |
+------+-------------------+----------+
| QSDF | 192.168.0.251:666 | unlocked |
+------+-------------------+----------+
```

kamel-rahim:


proxy/cs_actions.c
```diff
@@ -170,6 +170,11 @@ _registration (struct req_args_s *args, enum reg_op_e op, struct json_object *js
 		return _reply_format_error (args, BADREQ("Service type not specified"));
 	}
 
+	if (!validate_srvtype(si->type)) {
```
jfsmig: Here you fix more things than just "locking a service of an inexistant type".
You also check the service type when unlocking some service, and when pushing a service.
This is different than expected by the purpose of the PR, but why not (?), I don't succeed to find a cons.

PR #927, opened 2017-03-20
oio: Improve the behavior of download of contents with bad chunks

jfsmig:
Le's make the python API avoid chunks with dirty attributes (i.e. out of a quorum).

### DIFF
--- a/oio/api/object_storage.py
+++ b/oio/api/object_storage.py
@@ -139,15 +139,14 @@ def _sort_chunks(raw_chunks, ec_security):
             continue
         # sort by length
         bylength = byhash.values()
-        bylength.sort(lambda x, y: cmp(len(y), len(x)))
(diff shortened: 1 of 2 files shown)


oio/api/object_storage.py
```diff
@@ -125,6 +125,24 @@ def _sort_chunks(raw_chunks, ec_security):
...
+        for chunk in local_chunks:
+            h = chunk.get('hash')
+            if h not in byhash:
+                byhash[h] = list()
+            byhash[h].append(chunk)
+        if len(byhash) < 2:
+            continue
+        # sort by length
+        bylength = byhash.values()
+        bylength.sort(lambda x, y: cmp(len(y), len(x)))
```
fvennetier: `bylength.sort(key=len, reverse=True)` would be clearer.

PR #933, opened 2017-03-23
proxy: Add bulk delete content

kamel-rahim:
Permit to delete many content in one go for one container.
Modify the format of cli command 'openio object delete' to return a List.Lister.
Modify tests that depended on the old format response to use the new one.

The format of a request to the proxy is a json `{"contents":[{"name":"..."},{"name":"..."}]}`
and the url is /content/delete_many.
The format of the answer is a json `{"contents":[{"name":"...", "status":"...", "message":"..."}]}`


proxy/m2_actions.c
```diff
@@ -1886,6 +1886,77 @@ enum http_rc_e action_content_delete (struct req_args_s *args) {
...
+		return _reply_format_error(args,
+				BADREQ("Missing account or namespace"));
+
+	if (!json_object_object_get_ex(jbody, "contents", &jarray)
+			|| !json_object_is_type(jarray, json_type_array))
+		return _reply_format_error(args,
+				BADREQ("Invalid array of contents"));
+
+	GString *gresponse = g_string_sized_new(2048);
+	g_string_append(gresponse, "{\"contents\":[");
```
fvennetier: Since you `return` abruptly after checking input array, you should allocate `gresponse` after, or you will get a memory leak.

proxy/m2_actions.c
```diff
@@ -1886,6 +1886,77 @@ enum http_rc_e action_content_delete (struct req_args_s *args) {
...
+static enum http_rc_e
+_m2_content_delete_many (struct req_args_s *args, struct json_object * jbody) {
+	guint max_request = OIO_CONTENT_DELETE_MANY_NB_MAX_REQ;
+	GError *err = NULL;
+	const char *url_path = oio_url_get(args->url, OIOURL_PATH);
+	json_object *jarray = NULL;
+	PACKER_VOID(_pack) { return m2v2_remote_pack_DEL (args->url); }
+
+	if (!oio_url_get(args->url, OIOURL_ACCOUNT)
+			|| !oio_url_get(args->url, OIOURL_NS))
```
jfsmig: You don't need to check for the NS presence. The URL parsing already does it (NS present and expected)

proxy/m2_actions.c
```diff
@@ -1886,6 +1886,77 @@ enum http_rc_e action_content_delete (struct req_args_s *args) {
...
 
+static enum http_rc_e
+_m2_content_delete_many (struct req_args_s *args, struct json_object * jbody) {
+	guint max_request = OIO_CONTENT_DELETE_MANY_NB_MAX_REQ;
+	GError *err = NULL;
+	const char *url_path = oio_url_get(args->url, OIOURL_PATH);
+	json_object *jarray = NULL;
+	PACKER_VOID(_pack) { return m2v2_remote_pack_DEL (args->url); }
+
+	if (!oio_url_get(args->url, OIOURL_ACCOUNT)
```
jfsmig: You can consider using `oio_url_has_fq_container()` to check that all the container's fields are present

proxy/m2_actions.c
```diff
@@ -1886,6 +1886,77 @@ enum http_rc_e action_content_delete (struct req_args_s *args) {
...
+
+		if (!json_object_object_get_ex(jcontent, "name", &jname)
+				|| !json_object_is_type(jname, json_type_string)) {
+			g_string_append(gresponse, "\"(null)\",");
+			_append_status(gresponse, CODE_BAD_REQUEST,
+				 "Bad \"name\" field");
+		} else {
+			name = json_object_get_string(jname);
+			oio_url_set(args->url, OIOURL_PATH, name);
+			g_string_append_printf(gresponse, "\"%s\",", name);
```
jfsmig: Please escape the strings you write into the JSON messages. Consider using `oio_str_gstring_append_json_string()`, written by @fvennetier for this purpose.

proxy/m2_actions.c
```diff
@@ -1886,6 +1886,77 @@ enum http_rc_e action_content_delete (struct req_args_s *args) {
...
+		g_string_append_c(gresponse, '}');
+
+		if (i < jarray_len -1)
+			g_string_append_c(gresponse, ',');
+	}
+	oio_url_set(args->url, OIOURL_PATH, url_path);
+	g_string_append(gresponse, "]}");
+
+	return _reply_success_json(args, gresponse);
+
```
jfsmig: Non-necessary trailing white space.

proxy/m2_actions.c
```diff
@@ -1886,6 +1886,77 @@ enum http_rc_e action_content_delete (struct req_args_s *args) {
...
+				_append_status(gresponse, HTTP_CODE_NO_CONTENT, "No Content");
+			}
+		}
+
+		g_string_append_c(gresponse, '}');
+
+		if (i < jarray_len -1)
+			g_string_append_c(gresponse, ',');
+	}
+	oio_url_set(args->url, OIOURL_PATH, url_path);
```
jfsmig: This doesn't make sense. The variable is initiated with a `const char*` directly coming from the URL. On the first call to `oio_url_set()` that overrides the path invalidates the pointer. You should copy the string, instead.

proxy/m2_actions.c
jfsmig: `fq` stands for "fully qualified"

proxy/m2_actions.c
jfsmig: .. but using the `fq` function will implicitely check for the namespace.

oio/api/object_storage.py
```diff
@@ -482,6 +482,23 @@ def object_delete(self, account, container, obj, headers={}):
...
+        params = self._make_params(account, container)
+        if 'X-oio-req-id' not in headers:
+            headers['X-oio-req-id'] = utils.request_id()
+        unformatted_data = list()
+        for obj in objs:
+            unformatted_data.append({'name': obj})
+        data = json.dumps({"contents": unformatted_data})
+        resp, body = self._request(
+            'POST', uri, data=data, params=params, headers=headers)
+        results = []
```
jfsmig: We should manage the case of too many items sent at once.

jfsmig: We should manage the case of `delete_many` not supported

proxy/m2_actions.c
```diff
@@ -1886,6 +1886,78 @@ enum http_rc_e action_content_delete (struct req_args_s *args) {
...
+	for (guint i = 0; i < jarray_len; i++) {
+		struct json_object * jname = NULL;
+		struct json_object * jcontent = NULL;
+		const gchar *name = NULL;
+		jcontent = json_object_array_get_idx(jarray, i);
+		g_string_append(gresponse, "{\"name\":");
+
+		if (!json_object_object_get_ex(jcontent, "name", &jname)
+				|| !json_object_is_type(jname, json_type_string)) {
+			g_string_append(gresponse, "\"(null)\",");
```
jfsmig: You can use the `null` json object (without parenthesis and quotes)

jfsmig: This should cause a reply with the global status `400 Bad request`

PR #946, opened 2017-03-31
rawx: lower memory usage (especially when writing) (fix #944)

### LINKED ISSUE #944: High memory consumption on rawx during writes
When writing data to disk, the rawx services are consuming a lot of RAM (reports shows around 400MB for 1GB chunk size).
There may be a problem on the way httpd manage its buffers.

fvennetier:


rawx-apache2/src/rawx_repo_core.c
```diff
@@ -120,7 +120,7 @@ _finalize_chunk_creation(dav_stream *stream)
 static dav_error *
 _write_data_crumble_UNCOMP(dav_stream *stream)
 {
-	if ( 1 != fwrite(stream->buffer, stream->bufsize, 1, stream->f)) {
+	if (fwrite(stream->buffer, stream->buffer_offset, 1, stream->f) != 1) {
```
jfsmig: The shape `long_action(with_parmaters, that_are_long, enough_to_be, wrapped) != result` is less readable especially when the function call needs several lines.

PR #958, opened 2017-04-05
Fix memleaks and reduce the memory ping-pong

jfsmig:


core/ext.c
```diff
@@ -165,16 +165,15 @@ void ** oio_ext_array_concat (void **t0, void **t1) {
 
 /** @private */
 struct oio_ext_local_s {
-	gchar *reqid;
 	GRand *prng;
-	gboolean is_admin;
+	guint8 is_admin;
+	gchar reqid[63];
```
fvennetier: If this magic number is some kind of attempt at memory alignment, there is probably a better solution. By the way, do we need 63 characters for a request id?

PR #959, opened 2017-04-06
rawx: Do not compute the MD5 of the chunk when using EC

jfsmig:
the hash of the chunk doesn't make sense, and it is saved nowhere.


jfsmig:
The Python API still relies on the chunk_hash (for no real purpose) and it is a change way to big for the small expected gain, in a milestone already overdue.

PR #966, opened 2017-04-07
Minor python optimizations, prevent segfault at sqliterepo service shutdown

fvennetier:


oio/common/client.py
```diff
@@ -47,5 +47,5 @@ def _direct_request(self, method, full_url, **kwargs):
 
     def _request(self, method, url, **kwargs):
         endpoint = self.endpoint
-        url = '/'.join([endpoint.rstrip('/'), url.lstrip('/')])
+        url = '/'.join((endpoint.rstrip('/'), url.lstrip('/')))
```
jfsmig: :+1:

PR #968, opened 2017-04-07
Avoid filling the RPC with useless fields

jfsmig:
When the RPC mode is not set to "admin", don't tell "admin=false"
because this is the default.

### DIFF
--- a/metautils/lib/metautils_macros.h
+++ b/metautils/lib/metautils_macros.h
@@ -97,7 +97,7 @@ License along with this library.
 #define NS_STATE_VALUE_SLAVE      "slave"
 #define NS_STATE_VALUE_STANDALONE "standalone"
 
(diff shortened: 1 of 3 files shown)


metautils/lib/metautils_macros.h
```diff
@@ -97,7 +97,7 @@ License along with this library.
 #define NS_STATE_VALUE_SLAVE      "slave"
 #define NS_STATE_VALUE_STANDALONE "standalone"
 
-#define NAME_MSGKEY_ADMIN_COMMAND "ADM_CMD"
+#define NAME_MSGKEY_ADMIN_COMMAND "ADM"
```
fvennetier: Breaking change.

jfsmig: \o/

PR #969, opened 2017-04-07
core: Fix a memleak around the local cached configuration

jfsmig:


proxy/metacd_http.c
```diff
@@ -931,7 +926,7 @@ grid_main_configure (int argc, char **argv)
 	g_strlcpy(nsinfo.name, cfg_namespace, sizeof(nsinfo.name));
 	nsinfo.chunk_size = 1;
 
-	ns_conf = oio_cfg_cache_create(30 * G_TIME_SPAN_SECOND);
+	struct oio_cfg_handle_s *ns_conf = oio_cfg_cache_create(30 * G_TIME_SPAN_DAY);
```
fvennetier: 30 days?!

jfsmig: Well done, this has been forgotten.
A delay makes no sense. Only the operator of the platform knows when to reload the cache (after a valid délivery of the config files).

jfsmig:
Currently at commit 0835414994b319c8b04f99ac2b1b7c05c69b4437, the service needs to be restarted.

PR #971, opened 2017-04-07
Avoid thread pools with reserved threads

jfsmig:


conf.json
```diff
@@ -312,7 +312,7 @@
 				"descr": "In the current server, sets how many threads are allowed to the stats server. Keep this value really small, 1 should be enough for most usages, and consider increasing it if you have clues that the management of internal metrics is the bottleneck.",
 				"def": "1", "min": 1, "max": "1 << 31 - 1" },
 
-			{ "type": "int", "name": "server_threadpool_max_workers",
+			{ "type": "int", "name": "server_threadpool_max_tcp",
 				"key": "server.pool.max_workers",
```
fvennetier: In my opinion, the variable should have a name similar to the configuration value.

PR #972, opened 2017-04-07
server: Prevent DDoS with early drop of old cnx/req 

jfsmig:


server/network_server.c
```diff
@@ -1121,6 +1126,19 @@ _cb_worker(struct network_client_s *clt, struct network_server_s *srv)
...
 	}
 
+	/* The event stayed *really* long in the queue of the thread pool.
+	 * Let's close the connection, and let the client retry it's request. */
+	if (clt->events & CLT_READ) {
+		const gint64 now = oio_ext_monotonic_time();
+		if (clt->time.evt_in < OLDEST(now, G_TIME_SPAN_SECOND)) {
+			GRID_INFO("SLOW fd %d peer %s delay %"G_GINT64_FORMAT"ms",
+					clt->fd, clt->peer_name,
+					(now - clt->time.evt_in) / 1000);
```
fvennetier: 1000? -> `G_TIME_SPAN_MILLISECOND`

fvennetier: Shouldn't we `WARN` or `NOTICE` in such case?

fvennetier: I agree that warnings may fill the logs. Maybe we could warn from another thread, at most once every N seconds, if the `gq_count_overloaded` counter increases.

jfsmig: In case of a DoS, we would flood the log itself. With a message whose level is too high, it is even hard to turn the message off. I such a circumstance, I tend to restrict WARN and NOTICE levels for rather "rare" messages, or on process basis (important options at start-up).

What about a INFO or DEBUG for each request, and a WARN or NOTICE with a summary of the errors during a given period ?

PR #978, opened 2017-04-26
Add meta2 and meta1 databases rebuilding scripts

sebastienlapierre:
Add 2 bash scripts, each dedicated to rebuild meta2/meta1 databases.


tools/oio-meta1-rebuilder.sh
```diff
@@ -0,0 +1,70 @@
...
+     *)
+	usage
+	exit 0
+      ;;
+  esac
+done
+
+
+#Get account list
+redis_bin=$(which redis-cli)
```
vdombrovski: You get the path to redis-cli here, yet you don't use it in the following line

sebastienlapierre: thx i fixed it

jfsmig:
Rebased in #981

PR #986, opened 2017-05-03
rawx: Improvement and fixes around the uploads of chunks

jfsmig:
* More secure rollback() of uploads
* Atomic chunk creation
* MD5 compared between the request and the local computation: only effective when both are present.
* Local MD5 computation can be siabled by a configuration directive `grid_checksum: on|off|smart` where `smart` activates it when meaningful (i.e. not when managing an EC/ chunk).

### DIFF
--- a/rawx-apache2/src/rawx_repo_core.c
+++ b/rawx-apache2/src/rawx_repo_core.c
@@ -842,7 +842,13 @@ retry:
 		ds->compress_checksum = checksum;
 	}
 
-	ds->md5 = g_checksum_new (G_CHECKSUM_MD5);
+	ds->md5 = NULL;
+	if (conf->checksum_mode == CHECKSUM_ALWAYS) {
+		ds->md5 = g_checksum_new (G_CHECKSUM_MD5);
+	} else if (conf->checksum_mode == CHECKSUM_SMART) {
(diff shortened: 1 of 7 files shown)


jfsmig:
Also fixes the sending of the HTTP trailer. Prior to this PR, the `Trailer:` header was malformed and not understood by the server.

rawx-apache2/src/rawx_repo_core.c
```diff
@@ -722,29 +745,48 @@ rawx_repo_stream_create(const dav_resource *resource, dav_stream **result)
...
+			if (!e)
+				goto retry;
 			return e;
 		}
-		DAV_DEBUG_REQ(resource->info->request, 0, "open(%s) failed : %s", ds->pathname, strerror(errno));
+		DAV_DEBUG_REQ(resource->info->request, 0, "open(%s) failed : %s",
+				ds->pathname, strerror(errno));
 		return server_create_and_stat_error(resource_get_server_config(resource), p,
-			MAP_IO2HTTP(rv), 0, "An error occurred while opening a resource.");
+				MAP_IO2HTTP(rv), 0, "Chunk creation error");
```
fvennetier: `rv` is always 0 before this statement.

PR #992, opened 2017-05-10
cli: Fix the behavior of `openio object create` upon an error

jfsmig:
The whole command shouldn't fail at the first error that occurs, if several contents have been specified.

### DIFF
--- a/oio/cli/storage/obj.py
+++ b/oio/cli/storage/obj.py
@@ -113,29 +113,44 @@ class CreateObject(ContainerCommandMixin, lister.Lister):
             f.seek(currpos)
             return total_size
 
+        any_error = False
         properties = parsed_args.property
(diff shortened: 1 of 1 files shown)


oio/cli/storage/obj.py
```diff
@@ -113,29 +113,44 @@ def get_file_size(f):
...
+            except KeyboardInterrupt:
+                results.append((name, 0, None, 'Interrupted'))
+                any_error = True
+                break
+            except Exception:
+                import traceback
+                print "Failed to upload", obj, "in", container
+                traceback.print_exc()
+                any_error = True
+                results.append((name, 0, None, 'Failed'))
```
fvennetier: I would have put the type or message of the exception in the fourth field.

PR #998, opened 2017-05-17
Central config, round 2

### LINKED ISSUE #974: Reload configuration on special signal
Since #969 (4.0.0), sqliterepo-based services and oio-proxy do not reload their local configuration automatically, and you have to restart them to do so. We would like to be able to trigger a configuration reload on a special signal, for example  `SIGHUP` (`SIGUSR1` and `SIGUSR2` are already used to control logging level).

jfsmig:
* [x] Some variables have been renamed for more clarity, some have been removed because they were redundant
* [x] Units can be used in the central oio/sds.conf files, depending on the type of the variable the unit can time-related or size-related.
* [x] The export of all the variables in now achieved with a single `__constructor__`, for less code.
* [x] The variables are not split into client/server but packaged with the library that uses them. So that the service should only expose the variables from the libraries they bind to
* [x] The limits in terms of file descriptors (max bases, max cnx active/passive) have been centralized (not on the CLI anymore)
* [x] `oio-proxy` reloads its configuraiton upon SIGHUP signals
* [x] `oio-meta*-server` and `oio-sqlx-server` reload their configuraiton upon SIGHUP signals
* [x] `oio-rdir-server` reloads its configuraiton upon SIGHUP signals
* [x] Make the tests pass
* [x] Fix #974 (the "special signal" will be SIGHUP)


core/var.c
```diff
@@ -170,10 +179,88 @@ _record_set(struct oio_var_record_s *rec, union oio_var_default_u v)
...
+#define Mega 1000LL * Kilo
+#define Giga 1000LL * Mega
+#define Tera 1000LL * Giga
+#define Peta 1000LL * Tera
+
+#define Kibi 1024LL
+#define Mibi 1024LL * Kibi
+#define Gibi 1024LL * Mibi
+#define Tibi 1024LL * Gibi
+#define Pibi 1024LL * Tibi
```
fvennetier: Actually, it is *mebi*, *tebi* and *pebi*.

proxy/metacd_http.c
```diff
@@ -932,41 +954,32 @@ grid_main_configure (int argc, char **argv)
...
-	for (GSList *l = config_paths; l ; l = l->next) {
-		if (!l->data)
-			continue;
-		struct oio_cfg_handle_s *cfg =
-			oio_cfg_cache_create_fragment(l->data);
-		oio_var_value_all_with_config(cfg, ns_name);
-		oio_cfg_handle_clean(cfg);
-	}
-
+	/* To work properly, the PROXY needs to known the URL of the conscience */
```
fvennetier: *know*

PR #1003, opened 2017-05-19
Fix segfault in sqliterepo

fvennetier:
This happens in a weird case where the peers saved in a base do not include the service currently serving the base...

### DIFF
--- a/sqlx/sqlx_service.c
+++ b/sqlx/sqlx_service.c
@@ -496,6 +496,8 @@ label_retry:
 			// If cache was enabled, we can retry without cache
 			if (!nocache) {
 				nocache = TRUE;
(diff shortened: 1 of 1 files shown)


sqlx/sqlx_service.c
```diff
@@ -496,6 +496,8 @@ _get_peers_wrapper(struct sqlx_service_s *ss, const struct sqlx_name_s *name,
 			// If cache was enabled, we can retry without cache
 			if (!nocache) {
 				nocache = TRUE;
+				// freed by filter_services_and_clean() but still not NULL
+				peers = NULL;
```
murlock: should be done in `filter_services_and_clean` instead

PR #1012, opened 2017-05-23
Reduce CPU usage when sending events, warn when reaching max queue size

fvennetier:
Buffered event handling function was called too often.

### DIFF
--- a/events/oio_events_queue_beanstalkd.c
+++ b/events/oio_events_queue_beanstalkd.c
@@ -243,6 +243,21 @@ _poll_out (int fd)
 	return NULL;
 }
 
+static void
+_flush_buffered(struct oio_events_queue_s *self, struct _queue_BEANSTALKD_s *q)
+{
+	gint avail =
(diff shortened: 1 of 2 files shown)


events/oio_events_queue_beanstalkd.c
```diff
@@ -243,22 +243,39 @@ _poll_out (int fd)
 	return NULL;
 }
 
+static void
+_flush_buffered(struct oio_events_queue_s *self, struct _queue_BEANSTALKD_s *q)
+{
+	gint avail =
+		oio_events_common_max_pending - g_async_queue_length(q->queue);
+	if (avail < (gint) oio_events_common_max_pending / 100) {
```
jfsmig: Why such a division?

fvennetier: Less than 1% free. We warn before it actually fails.

jfsmig: It might continuously vomit and fill the logs. But let's give it a try!

PR #1026, opened 2017-06-02
Account: send a lua script to redis to update containers.

kamel-rahim:
Replace the lock/unlock by a script to redis.
Since a script is atomic it should prevent the problem of tilted redis that left lock.

### DIFF
--- a/oio/account/backend.py
+++ b/oio/account/backend.py
@@ -4,6 +4,8 @@ import math
 
 import redis
 import redis.sentinel
+from werkzeug.exceptions import NotFound, Conflict
+from oio.common.exceptions import NoSuchAccount
 from oio.common.utils import Timestamp
 from oio.common.utils import int_value
 from oio.common.utils import true_value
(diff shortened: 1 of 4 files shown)


oio/account/backend.py
```diff
@@ -202,83 +204,115 @@ def update_container(self, account_id, name, mtime, dtime, object_count,
...
+            dtime = '0'
+        if object_count is None:
+            object_count = 0
+        if bytes_used is None:
+            bytes_used = 0
+
+        lua = (
+               # With lua float we are losing precision for this reason
+               # we keep the number as a string
+               " local is_sup = function(a,b)\n"
```
murlock: you should use
"""
script lua
"""
(but I don't check if conn.register_script accept string or only a tuple)

oio/account/backend.py
```diff
@@ -202,83 +204,115 @@ def update_container(self, account_id, name, mtime, dtime, object_count,
...
+               "end;\n"
+               )
+
+        keys = [account_id, AccountBackend.ckey(account_id, name),
+                ("containers:%s" % (account_id)),
+                ("account:%s" % (account_id))]
+        args = [name, mtime, dtime, object_count, bytes_used,
+                self.autocreate, Timestamp(time()).normal, EXPIRE_TIME]
+
+        update = conn.register_script(lua)
```
murlock: What is the cost of registering script for each update_container call  ?
Could it be cached on `self` lifetime ?

oio/account/backend.py
```diff
@@ -202,83 +288,31 @@ def update_container(self, account_id, name, mtime, dtime, object_count,
...
+        args = [name, mtime, dtime, object_count, bytes_used,
+                self.autocreate, Timestamp(time()).normal, EXPIRE_TIME]
         try:
-            pipeline = conn.pipeline(True)
-            pipeline.hmset(AccountBackend.ckey(account_id, name), record)
-            if deleted:
-                pipeline.expire(AccountBackend.ckey(account_id, name),
-                                EXPIRE_TIME)
-                pipeline.zrem('containers:%s' % account_id, name)
+            self.script_update_container(keys=keys, args=args)
```
fvennetier: You have to pass `client=self.conn` because the client that created the script may no longer be available (especially if using Redis Sentinel).

PR #1039, opened 2017-06-09
ContentRebuilderFilter: Merge with NotifyFilter (fix #1033)

### LINKED ISSUE #1033: ContentRebuildFilter is too similar to NotifyFilter
These two classes have to be merged. Only the default tube is different. By properly configuring `NotifyFilter` you can achieve the same functionality as `ContentRebuildFilter`.

kamel-rahim:


oio/event/filters/notify.py
```diff
@@ -2,19 +2,27 @@
...
 
     def process(self, env, cb):
         data = json.dumps(env)
+        event = Event(env)
         try:
-            self.beanstalk.put(data)
+            if event.event_type == EventTypes.CONTENT_BROKEN:
+                self.beanstalk_rebuild.put(data)
+            else:
+                self.beanstalk_notify.put(data)
```
fvennetier: These modifications are useless. If notifications are needed at the same time as rebuild, there will be two instances of this class, each one writing in its own tube.

PR #1044, opened 2017-06-15
Check validity of oio_url_s instance

murlock:
Only core, meta2 and proxy are now protected (I hope, please correct me if I miss something).

Path and version are always checked with oio_url_check.
Namespace is only checked if current one is passed as argument.

### DIFF
--- a/core/url.c
+++ b/core/url.c
@@ -506,3 +506,29 @@ oio_url_to_json (GString *out, struct oio_url_s *u)
 	}
 }
 
+gboolean
+oio_url_check(const struct oio_url_s *u, const char *namespace, const gchar **err )
+{
+#define _ERR(v)  \
+    if (err) { \
+        *err = v; \
+    }
+
+    _ERR(NULL);
+    if (namespace && u->ns && strcmp(namespace, u->ns)) {
+        _ERR("namespace");
+        return 0;
+    }
+    if (u->version && !oio_str_is_number(u->version, NULL)) {
+        _ERR("version");
+        return 0;
+    }
+
+    if (u->path && !g_utf8_validate(u->path, -1, NULL)) {
+        _ERR("path");
+        return 0;
+    }
+    return 1;
+
+#undef _ERR
+}
--- a/meta1v2/meta1_gridd_dispatcher.c
+++ b/meta1v2/meta1_gridd_dispatcher.c
@@ -50,9 +50,8 @@ static GByteArray *encode_and_clean(GByteArray* (*e)(gchar**), gchar **pv) {
 
 static gboolean
 meta1_dispatch_v2_USERCREATE(struct gridd_reply_ctx_s *reply,
-		struct meta1_backend_s *m1, gpointer ignored UNUSED)
+		struct meta1_backend_s *m1, struct oio_url_s *url)
 {
-	struct oio_url_s *url = metautils_message_extract_url (reply->request);
 	reply->subject("%s|%s", oio_url_get(url, OIOURL_WHOLE), oio_url_get(url, OIOURL_HEXID));
 
 	gsize length = 0;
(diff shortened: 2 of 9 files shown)


meta2v2/meta2_filters_extract.c
```diff
@@ -77,6 +77,12 @@ meta2_filter_extract_header_url(struct gridd_filter_ctx_s *ctx,
 {
 	TRACE_FILTER();
 	struct oio_url_s *url = metautils_message_extract_url (reply->request);
+    if (!oio_url_check(url)) {
+		meta2_filter_ctx_set_error(ctx, NEWERROR(CODE_BAD_REQUEST,
+					"Invalid request, Empty / Invalid body"));
```
fvennetier: The message should tell the problem comes from the content/container URL.

proxy/metacd_http.c
```diff
@@ -238,7 +238,9 @@ handler_action (struct http_request_s *rq, struct http_reply_ctx_s *rp)
 		if (oio_url_has(args.url, OIOURL_VERSION) && !oio_str_is_number(
 					oio_url_get(args.url, OIOURL_VERSION), &args.version))
 			rc = _reply_format_error(&args, BADREQ("Invalid version"));
-		else {
+		else if (!oio_url_check(url)) {
+			rc = _reply_format_error(&args, BADREQ("Invalid parameters"));
```
fvennetier: Same here, "parameters" is too vague.

core/url.c
```diff
@@ -506,3 +506,8 @@ oio_url_to_json (GString *out, struct oio_url_s *u)
 	}
 }
 
+gboolean
+oio_url_check(const struct oio_url_s *u)
+{
+    return !u->path || g_utf8_validate(u->path, -1, NULL);
```
jfsmig: You could also validate the format of the version, e.g. with `oio_str_is_number()`.

proxy/metacd_http.c
```diff
@@ -238,7 +238,9 @@ handler_action (struct http_request_s *rq, struct http_reply_ctx_s *rp)
 		if (oio_url_has(args.url, OIOURL_VERSION) && !oio_str_is_number(
```
jfsmig: Checking the version could be part of the `oio_url_check()` function.

proxy/metacd_http.c
```diff
@@ -238,7 +238,9 @@ handler_action (struct http_request_s *rq, struct http_reply_ctx_s *rp)
...
 					oio_url_get(args.url, OIOURL_VERSION), &args.version))
 			rc = _reply_format_error(&args, BADREQ("Invalid version"));
-		else {
```
jfsmig: You could also check the NS (if present) is managed by the current `oio-proxy`, as in https://github.com/open-io/oio-sds/blob/master/proxy/lb_actions.c#L229

meta1v2/meta1_gridd_dispatcher.c
```diff
@@ -46,13 +46,29 @@ static GByteArray *encode_and_clean(GByteArray* (*e)(gchar**), gchar **pv) {
...
 /* -------------------------------------------------------------------------- */
 
 static gboolean
 meta1_dispatch_v2_USERCREATE(struct gridd_reply_ctx_s *reply,
 		struct meta1_backend_s *m1, gpointer ignored UNUSED)
 {
-	struct oio_url_s *url = metautils_message_extract_url (reply->request);
+	struct oio_url_s *url = _extract_url(reply);
+	if (!url) {
+		return TRUE;
```
jfsmig: It is highly recommended either to reply something to the client or to abruptly close the connection with a `return FALSE`.

murlock: I bet that my function _extract_url is misnamed because a reply is done with BADREQ

PR #1051, opened 2017-06-23
oio/api: allow append operations

murlock:
This patch introduces a new parameter for object_create to support
appending data to an existing object.

If object doesn't exist, it is created, only if container exists as
flag append can not be used at same time with flag autocreate.

### DIFF
--- a/oio/common/client.py
+++ b/oio/common/client.py
@@ -52,11 +52,5 @@ class ProxyClient(HttpApi):
             headers["X-oio-action-mode"] = "autocreate"
             kwargs = kwargs.copy()
             kwargs.pop("autocreate")
-        if kwargs.get("append"):
-            if not headers:
-                headers = dict()
-            headers["X-oio-action-mode"] = "append"
-            kwargs = kwargs.copy()
-            kwargs.pop("append")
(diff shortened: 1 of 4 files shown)


oio/common/client.py
```diff
@@ -52,5 +52,11 @@ def _direct_request(self, method, url, session=None, headers=None,
             headers["X-oio-action-mode"] = "autocreate"
             kwargs = kwargs.copy()
             kwargs.pop("autocreate")
+        if kwargs.get("append"):
```
fvennetier: Append mode should not be selected by a header, but preferably by a query string parameter.

oio/container/client.py
```diff
@@ -310,7 +310,7 @@ def content_create(self, account=None, reference=None, path=None,
         if chunk_method is not None:
             hdrs['x-oio-content-meta-chunk-method'] = chunk_method
         resp, body = self._direct_request(
-            'POST', uri, data=data, params=params, autocreate=True,
+            'POST', uri, data=data, params=params,
```
fvennetier: You should add `append` parameter to the method signature, and then set it in `params`.

PR #1058, opened 2017-06-27
support restoration of container by range

murlock:
When a dump operation is launched, the Manifest is stored as first file of the archive.

Please note that restoration by range must be a multiple of 1 MB, as empty block may be inserted to avoid header split.

When untaring a dump on Linux filesystem, you should provide option --ignore-zeroes to avoid 'A lone zero block at %d' error.

### DIFF
--- a/oio/container/app.py
+++ b/oio/container/app.py
@@ -341,8 +341,8 @@ class ContainerStreaming(RedisConn):
 
         self.proxy = ObjectStorageApi(self.conf.get("namespace", NS))
         self.url_map = Map([
-            Rule('/v1.0/dump', endpoint='dump'),
-            Rule('/v1.0/restore', endpoint='restore'),
+            Rule('/v1.0/container/dump', endpoint='dump'),
+            Rule('/v1.0/container/restore', endpoint='restore'),
         ])
         self.logger = get_logger(self.conf, name="ContainerStreaming")
         super(ContainerStreaming, self).__init__(self.conf)
--- a/oio/container/app.py
+++ b/oio/container/app.py
@@ -22,12 +22,12 @@ try:
 except ImportError:
     import json  # noqa
 
+from collections import OrderedDict
 import math
 import re
 import os
 from tarfile import TarInfo, REGTYPE, NUL, PAX_FORMAT, BLOCKSIZE, XHDTYPE, \
                     DIRTYPE, AREGTYPE
-import time
 
 from werkzeug.wrappers import Request, Response
 from werkzeug.routing import Map, Rule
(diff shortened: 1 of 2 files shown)


oio/container/app.py
```diff
@@ -519,6 +541,57 @@ def on_dump(self, req):
...
+        append = False
+        mode = "full"
+
+        if req.headers.get('range'):
+            _, _, start_block, end_block = self._extract_range(req,
+                                                               blocks=None)
+            append = True
+            mode = "range"
+
+            cur_state = self.conn.get("restore-%s-%s" % (account,
```
fvennetier: Every other Redis-backed OpenIO service (oiofs, account) uses `:` as a separator (don't ask me why).

oio/container/app.py
```diff
@@ -519,6 +541,57 @@ def on_dump(self, req):
...
+                    [x['end_block'] for x in cur_state['manifest']]) + 1
+            else:
+                if not cur_state:
+                    raise UnprocessableEntity("First segment is not available")
+                cur_state = json.loads(cur_state)
+
+                if start_block != cur_state['end']:
+                    raise UnprocessableEntity(
+                        "Segment was already written "
+                        "or an error occurs previously")
```
fvennetier: *has occured*

oio/container/app.py
```diff
@@ -376,11 +380,24 @@ def generate_oio_map(self, account, container):
...
+                    ((start_block + tar.header_blocks) / self.CHUNK):
+                # header on boundary, we have to add empty block
+                padding = self.CHUNK - divmod(start_block, self.CHUNK)[1]
+                map_objs.append({
+                    'block': padding,
+                    'size': padding * BLOCKSIZE,
+                    'start_block': start_block,
+                    'slo': False,
+                    'hdr_block': padding,
+                    'end_block': start_block + padding - 1
```
murlock: I will rename them in dedicated commit

oio/container/app.py
```diff
@@ -424,7 +441,8 @@ def _do_head(self, _, account, container):
             'X-Blocks': sum([i['block'] for i in results]),
             'Content-Length': sum([i['block'] for i in results]) * BLOCKSIZE,
             'Accept-Ranges': 'bytes',
-            'Content-Type': 'application/tar'
+            'Content-Type': 'application/tar',
+            'X-Manifest-Container': json.dumps(results),
```
murlock: Agreed, I will store the manifest as special file like it was done for container properties to avoid this kind of issue.

oio/container/app.py
murlock: Dine

oio/container/app.py
murlock: Done

oio/container/app.py
```diff
@@ -52,25 +52,25 @@
 
 NS = os.getenv("OIO_NS")
 CONTAINER_PROPERTIES = ".container_properties"
+MANIFEST_PROPERTIES = ".manifest_properties"
```
fvennetier: `CONTAINER_MANIFEST = ".container_manifest"`.

oio/container/app.py
fvennetier: It's so confusing that I misunderstood.
* `start_block` is the index of the first block of the entry.
* `end_block` is the index of the last block of the entry.
*`hdr_block` is the number (!) of blocks required to store the header of the entry.
*`block` is the number (!) of blocks required to store the entry (including the header blocks).

Am I right?

murlock: Yes

PR #1064, opened 2017-06-29
Improve the behavior with huge amounts of elections

jfsmig:
* [x] Add a tool to stress a ZK cluster: know its limit
* [x] Shuffle the ZK endpoints in a connection string: for a better balancing of the connections among the nodes
* [x] Allow to balance a connection to a ZK cluster over several ZK handles (and as many connections): for smaller transactions, then less jute.maxbuffer errors
* [x] Allow several ZK cluster in the configuration: when the number of bases is too high


sqlx/sqlx_service.c
```diff
@@ -359,35 +359,82 @@ _configure_peering (struct sqlx_service_s *ss)
...
+}
+
+static gboolean
+_configure_multi_zk(struct sqlx_service_s *ss,
+		guint mux_factor, const char *realprefix, const char *zk_url)
+{
+	if (mux_factor <= 0)
+		return FALSE;
+
+	gchar **shards = g_strsplit(zk_url, ";", -1);
```
fvennetier: `OIO_CSV_SEP2`

PR #1081, opened 2017-07-07
Backup/Restore optimization

murlock:
- Remove list of blocks to only keep its boundaries
- Use object_create with a LimitedStream to avoid wasting RAM

### DIFF
--- a/oio/container/backup.py
+++ b/oio/container/backup.py
@@ -146,16 +146,16 @@ class ContainerTarFile(object):
     """ Expose a File Object API to be used with wrap_file """
 
     def __init__(self, storage_api, account, container,
-                 blocks, oio_map, logger):
+                 ranges, oio_map, logger):
         self.acct = account
         self.container = container
-        self.blocks = blocks
+        self.ranges = ranges
         self.oio_map = oio_map
         self.manifest = oio_map[:]
         self.storage = storage_api
         self.logger = logger
-        if not blocks:
-            self.logger.warn('no blocks provided for %s %s', account,
+        if len(ranges) != 2:
+            self.logger.warn('no valid ranges provided for %s %s', account,
                              container)
 
     def __iter__(self):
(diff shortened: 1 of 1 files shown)


oio/container/backup.py
```diff
@@ -319,17 +320,16 @@ def read(self, size=-1):
...
                 end_block = val['end_block']
 
-            # FIXME: should be done in same loop
-            blocks = [x for x in self.blocks if x <= end_block]
-            # remove selected from globals list
-            _c = set(blocks)
-            self.blocks = [x for x in self.blocks if x not in _c]
-            # shift selected blocks to object
+            assert self.blocks[0] >= val['start_block']
+            assert self.blocks[0] <= self.blocks[1], "Got start %d / end %d" % (self.blocks[0], self.blocks[1])
```
jfsmig: The line is too long, this cannot pass the PEP8 compliance tests.

oio/container/backup.py
```diff
@@ -555,10 +555,8 @@ def _do_get(self, req, account, container):
         length = blocks * BLOCKSIZE
 
         if 'Range' not in req.headers:
-            # TODO: instead expanding blocks to a full list,
-            # use [first_block - last_block]
             tar = ContainerTarFile(self.proxy, account, container,
-                                   range(blocks), results, self.logger)
+                                    [0, blocks-1], results, self.logger)
```
jfsmig: Not visually aligned, this cannot pass the PEP8 compliance test.

oio/container/backup.py
```diff
@@ -721,11 +717,14 @@ def read(size):
                         kwargs['mime_type'] = hdrs['mime_type']
                         del hdrs['mime_type']
 
-                    self.proxy.object_create(account, container,
-                                             obj_name=inf.name,
-                                             append=append,
-                                             data=read(inf.size),
-                                             **kwargs)
+                    ret = self.proxy.object_create(account, container,
```
jfsmig: That method is documented as returning a `list` of chunks, size and hash of the what has been uploaded.
Why not assign a tuple is the return of the method? It would be clearer.
And if the function ever returns something different (without raising any exception), it is an error.

oio/container/backup.py
```diff
@@ -154,6 +154,7 @@ def __init__(self, storage_api, account, container,
         if not blocks:
             self.logger.warn('no blocks provided for %s %s', account,
                              container)
+        assert len(blocks) == 2
```
fvennetier: `ranges` is weird since this is just ONE range.

PR #1104, opened 2017-07-27
Python API: handle ''ServiceBusy" exceptions returned by oio-proxy

AymericDu:
Catch error 503 from proxy and retry with exponential backoff.
If the proxy is still busy, throws the `ServiceBusy` exception.

### DIFF
--- a/oio/common/client.py
+++ b/oio/common/client.py
@@ -2,6 +2,12 @@ from oio.common.utils import get_logger
 from oio.common.utils import load_namespace_conf
 from oio.common.utils import validate_service_conf
 from oio.api.base import HttpApi
+from oio.common.exceptions import ServiceBusy
+from random import randrange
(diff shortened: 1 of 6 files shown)


oio/common/client.py
```diff
@@ -52,5 +71,18 @@ def _direct_request(self, method, url, headers=None,
...
+                return super(ProxyClient, self)._direct_request(
+                    method, url, headers=headers, **kwargs)
+            except ServiceBusy:
+                if i >= self._attempts_request - 1:
+                    raise
+                # retry with exponential backoff
+                ProxyClient._waitTimeExp(i + 1)
+
+    @staticmethod
+    def _waitTimeExp(attempts):
```
fvennetier: Should be snake case, not camel case. Something like `_exp_sleep`.

PR #1145, opened 2017-08-25
oio/backup: fix invalid restore

### LINKED ISSUE #1143: Object blocks misplaced on restore
With __c11__ being the original container, and __c22__ and __c23__ being the restored containers, the result is as follows:

image

The iso files can be obtained from the debian website. This was done using the latest commit that affected the `backup.py` file (i.e. 897d41b)

- First grab the backup file using curl:
`curl -o backup.tar "http://ip:port/v1.0/container/dump?acct=testacc&ref=c11"`
- Then split the file into 1073741824 chunks:
`split -b 1073741824 backup.tar`
- Curl the chunks to the c22 container (The URLs were omitted)
```
curl -XPUT -H "Range: bytes=0-1073741823" --data-binary @xaa 
curl -XPUT -H "Range: bytes=1073741824-2147483647" --data-binary @xab 
curl -XPUT -H "Range: bytes=2147483648-3221225471" --data-binary @xac 
curl -XPUT -H "Range: bytes=3221225472-4294967295" --data-binary @xad 
curl -XPUT -H "Range: bytes=4294967296-5368709119" --data-binary @xae 
curl -XPUT -H "Range: bytes=5368709120-6442450943" --data-binary @xaf 
curl -XPUT -H "Range: bytes=6442450944-7516192767" --data-binary @xag 
curl -XPUT -H "Range: bytes=7516192768-8589934591" --data-binary @xah 
curl -XPUT -H "Range: bytes=8589934592-9663676415" --data-binary @xai 
curl -XPUT -H "Range: bytes=9663676416-10737418239" --data-binary @xaj 
curl -XPUT -H "Range: bytes=10737418240-11811160063" --data-binary @xak 
curl -XPUT -H "Range: bytes=11811160064-12884901887" --data-binary @xal 
curl -XPUT -H "Range: bytes=12884901888-13164868607" --data-binary @xam
```
 _(At this point no erros are yielded by the API)_

It should also be noted that when the `8589934592-9663676415` range is successfully PUT, the backup API stops appending to the second file and move on to the third.

- cURL the whole backup to the c23 container
`curl -XPUT --data-binary @backup.tar "http://ip:port/v1.0/container/restore?acct=testacc&ref=c23"`

And then the output is as shown in the picture above. 

A sample of the `restore:testacc:c22` redis value:
```
> GET restore:testacc:c22
"{\"end\": 18874368, \"last_block\": 25712634, \"manifest\": [{\"blocks\": 5, \"end_block\": 4, \"hdr_blocks\": 1, \"name\": \".__oio_container_manifest\", \"size\": 597, \"slo\": null, \"start_block\": 0}, {\"blocks\": 7431075, \"end_block\": 7431079, \"hdr_blocks\": 3, \"name\": \"debian-9.0.0-amd64-DVD-1.iso\", \"size\": 3804708864, \"slo\": null, \"start_block\": 5}, {\"blocks\": 9156407, \"end_block\": 16587486, \"hdr_blocks\": 3, \"name\": \"debian-9.0.0-amd64-DVD-2.iso\", \"size\": 4688078848, \"slo\": null, \"start_block\": 7431080}, {\"blocks\": 9125147, \"end_block\": 25712633, \"hdr_blocks\": 3, \"name\": \"debian-9.0.0-amd64-DVD-3.iso\", \"size\": 4672073728, \"slo\": null, \"start_block\": 16587487}], \"start\": 16777216}"
```

murlock:
A regression was introduced during f616a41ec as same var size was
used to store HTTP Content-Length and object uploaded to SDS.

This commit fix that by introducing new variable name (req_size) for
HTTP Content-Length and enforce that total data consumed are now
same size as HTTP.

It also enforce test to ensure all objects put are really restored.

This PR will closes #1143

### DIFF
--- a/oio/container/backup.py
+++ b/oio/container/backup.py
@@ -769,7 +769,8 @@ class ContainerBackup(RedisConn, WerkzeugApp):
                         file_or_path=LimitedStream(req.stream, inf.size),
                         **kwargs)
                     if size != inf.size:
-                        raise UnprocessableEntity()
+                        raise UnprocessableEntity(
+                            "Object created is smaller than expected")
                     r['consumed'] += size
                     if hdrs:
                         self.proxy.object_set_properties(account, container,
@@ -782,7 +783,8 @@ class ContainerBackup(RedisConn, WerkzeugApp):
                 read(BLOCKSIZE - inf.size % BLOCKSIZE)
 
         if req_size != r['consumed']:
-            raise UnprocessableEntity()
+            raise UnprocessableEntity(
+                "Invalid length of data consumed by restoration")
 
         if mode == 'full' or end_block == cur_state['last_block']:
             code = 201
--- a/oio/container/backup.py
+++ b/oio/container/backup.py
@@ -648,7 +648,7 @@ class ContainerBackup(RedisConn, WerkzeugApp):
     @redis_cnx
     def _do_put(self, req, account, container):
         """Manage PUT method for restoring a container"""
-        size = int(req.headers['content-length'])
+        req_size = int(req.headers['content-length'])
         start_block, end_block = (None, None)
         append = False
         mode = "full"
@@ -694,7 +694,7 @@ class ContainerBackup(RedisConn, WerkzeugApp):
                         offset = (start_block - entry['start_block']
                                   - entry['hdr_blocks'])
                         inf.size = entry['size'] - offset * BLOCKSIZE
-                        inf.size = min(inf.size, size)
+                        inf.size = min(inf.size, req_size)
                         break
                     raise UnprocessableEntity('Header is broken')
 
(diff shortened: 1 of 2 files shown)


oio/container/backup.py
```diff
@@ -781,6 +781,9 @@ def read(size):
             if inf.size % BLOCKSIZE:
                 read(BLOCKSIZE - inf.size % BLOCKSIZE)
 
+        if req_size != r['consumed']:
+            raise UnprocessableEntity()
```
aneutron: While this detects the error and fails, it would probably be better to specifiy in the returned UnprocessableEntity why it failed. For example a simple `Consumed size and expected size differ` would do the trick. _(Too many unprocessable entities are returned in the API, and it can be confusing to debug)_

murlock: Indeed, I've added a small sentence for both empty UnprocessableEntity

PR #1166, opened 2017-09-08
Sqliterepo: add check of 'when_unstable' for every election state

kamel-rahim:


sqliterepo/election.c
```diff
@@ -2203,6 +2203,10 @@ member_action_to_LISTING(struct election_member_s *member)
 		return member_leave_on_error(member, zrc);
 	}
 
+	// if comming from a stable state like SLAVE
+	if (!member->when_unstable)
```
jfsmig: :+1:

PR #1175, opened 2017-09-20
[WIP] Prepare the sharding of containers (step5: meta2)

jfsmig:
* A marker is now used, implemented with a flag in the admin table, to note that a container has ever been full. That marker cannot be removed the clean way, with meta2 requests.
* The  client of a meta2 service now marks its requests with a flag: is the base known to be the last (the last base has never been full --no flag-- while the previous have the flag). If it is, the "FULL" marker must not be present (and vice versa). If that condition is not fulfilled, an error with code=CODE_SHARD_FULL is generated, to tell the client to refresh the list of bases it knows.
* The checks have been added in the meta2 backend, at any the opening of a container.
* The check is avoided when the container is opened with the "LOCAL" flag (admin purposes)
* A notion is introduced: a "meta2 operation target" gathering a container URL, the sequence number and flags common to all the requests (last, master only, local).
* The sequence number is now mandatory for the backend operations, and set to the default value of 1, for backward compliance purposes.
* the dispatcher of requests has been substantially simplified to get rid of the confusing logic of filters.
* As a remarkable side change, when generating beans/chunks, there is no check anymore on the possible presence of an alias with the same name. This is a legacy piece of logic, unused, obsolete.

TODO:
* Ensure the backward compliance around the flags on meta2 requests.


jfsmig:
À un moment, il faut se faire une raison...

PR #1178, opened 2017-09-21
rawx: accept APR_INCOMPLETE status for more cases

kamel-rahim:


racciari:
This patch fixes my problem on ARM ;-)

PR #1179, opened 2017-09-21
CLI: Add commands to debug elections

AymericDu:


oio/cli/election/election.py
```diff
@@ -0,0 +1,126 @@
...
+        json = jsonlib.dumps(json, indent=4, sort_keys=True)
+    return json
+
+
+class PingElection(lister.Lister):
+    """Trigger or refresh an election"""
+
+    log = getLogger(__name__ + '.PingElection')
+
+    def get_parser(self, prog_name):
```
fvennetier: All commands of this class take the same parameters. You should consider creating a function (or a mixin class, like in [`oio/cli/object/object.py`](https://github.com/open-io/oio-sds/blob/4.x/oio/cli/object/object.py#L23)) to create the parser. By the way, `reference` is not mandatory, you can interact with an election knowing only the CID.

oio/cli/election/election.py
```diff
@@ -0,0 +1,126 @@
...
+        account = self.app.client_manager.account
+        reference = parsed_args.reference
+        service_type = parsed_args.service_type
+
+        data = self.app.client_manager.election.election_status(
+            account=account, reference=reference, service_type=service_type)
+
+        data["peers"] = format_json(parsed_args, data["peers"])
+
+        return zip(*sorted(data.iteritems()))
```
fvennetier: Here you should probaly return a listing.

PR #1191, opened 2017-10-02
Fix small typos & english glitches

vincent-legoll:


vincent-legoll:
What's the matter with this Travis test failing ?

jfsmig:
The travis-ci infrasturcture is sometimes really ... slow.
Some tests depend on (decently high and fair enough) timeouts, even if this is a bad idea in general to depend on time in the context of the tests.
In this case, the python `tests.functional.rdir.test_server.TestRdirServer2` starts a sandboxed rdir, and waits for its plain startup (or the listen socket appears, or the timeout is raised). Here, the rdir didn't start fast enough.

openio-private:
Yeah, thanks for the explanation, I was expecting something along those lines...

PR #1206, opened 2017-10-04
proxy,conscience-agent: Fix the verb of the /forward/stats route

jfsmig:


jfsmig:
@fvennetier: The purpose of the PR is to fix a temporary "hard to upgrade" situation that came with 

Prior to the commit 9fcc0b99fa1f51694bd4085cfd1f31848a1422ab, we had both GET and POST managed:
```
SET("/forward/stats/#GET", action_forward_stats);
SET("/forward/stats/#POST", action_forward_stats);
```

But the commit changed this and allowed only GET, while the conscience-agent is still using POST:
```
SET("/forward/stats/#GET", action_forward_stats);
```

We now restore the GET+POST, with a mention to remove the POST that makes no sense, as soon as the platform has been upgraded:
```
SET("/forward/stats/#GET", action_forward_stats);
/* TODO(jfs): remove in a further release, present for the sake of backward
 * compliance and smooth transition */
SET("/forward/stats/#POST", action_forward_stats);
```

PR #1208, opened 2017-10-06
Speed the exit process of the conscience-agent

jfsmig:
During the CI tests, especially when the load is high on large deployments, the tests sometimes fail because the conscience-agent is too slow to exit. It may even take several minutes to do so. It appears it remains sticked on loop-failing HTTP requests to the oio-proxy (that is down), retried periodically.

In this context, we:
* Allowed the proxy to react the same way to more exit signals
* Broke a main "while True" loop with the help of a flag, unset when an exit signal is received
* Forbid urllib3 to do retries when registering services

### DIFF
--- a/oio/common/daemon.py
+++ b/oio/common/daemon.py
@@ -35,16 +35,33 @@ class Daemon(object):
     def run(self, *args, **kwargs):
         raise NotImplementedError('run not implemented')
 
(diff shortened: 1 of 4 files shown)


jfsmig:
Thanks to @kamel-rahim for the diagnosis

PR #1216, opened 2017-10-17
sqlx,meta*: Split the main background task of re-kicking elections

jfsmig:
Several tasks with each one:
* its own critical section on the election set
* its trigger period (configurable)
* its threshold delay to warn for a long lock (configurable)

Now we should know when the lock is held for too long, and also split the critical section to remain fair with the other threads.

Also, the limit of elections that receive an event each turn is not limited any more. This was a sign of an ugly workaround to prevent a long-lock.


conf.json
```diff
@@ -364,10 +364,35 @@
...
+				"def": "5", "min": 0, "max": "86400" },
+
+			{ "type": "uint", "name": "sqliterepo_election_task_TIMER_period",
+				"key": "sqliterepo.election.task.timer.period",
+				"descr": "In jiffies, how often the elections waiting for timers are fired",
+				"def": "1", "min": 0, "max": "86400" },
+
+			{ "type": "uint", "name": "sqliterepo_election_task_PING_period",
+				"key": "sqliterepo.election.task.ping.period",
+				"descr": "In jiffies, how often a PING is maybe triggered for elections in a final state.",
```
vincent-legoll: s/is maybe/may be/

conf.json
```diff
@@ -364,10 +364,35 @@
...
+				"def": "15", "min": 0, "max": "86400" },
+
+			{ "type": "monotonic", "name": "sqliterepo_election_task_EXIT_alert",
+				"key": "sqliterepo.election.task.exit.alert",
+				"descr": "When NONE elections are expired, report a warning if the background task holds the lock longer than this value.",
+				"def": "200ms", "min": 0, "max": "max" },
+
+			{ "type": "monotonic", "name": "sqliterepo_election_task_TIMER_alert",
+				"key": "sqliterepo.election.task.timer.alert",
+				"descr": "When timer are raised on elections, report a warning if the background task holds the lock longer than this value.",
```
vincent-legoll: s/timer are/timers are/

PR #1232, opened 2017-10-23
oio: Comply to the latest PEP8 rules enforced by flake8

jfsmig:


oio/cli/clientmanager.py
```diff
@@ -117,15 +117,15 @@ def get_flatns_manager(self):
         bitlength, offset, size = None, 0, 0
         try:
             bitlength = int(options['flat_bitlength'])
-        except:
+        except Exception:
```
vincent-legoll: I think pylint still warns about a "too broad except clause" for this.
Couldn't we just catch the relevant ones and tell the user specific error messages to help them fix their config ?

fvennetier: pylint complaints, but fortunately it's only a warning (not an error). And besides pylint is not enforced in our test suites.

Issue #1087, opened 2017-07-18
Use same HTTP lib for all chunk operations


fvennetier:
A portion of `oio.blob.clien.BlobClient` uses **requests**, whereas the other portion uses `oio.api.io.ChunkReader`, which uses eventlet's **httplib**. This is not consistent. We would like to get rid of **requests**.

fvennetier:
`tests.utils.BaseTestCase` also uses **requests**.

PR #1238, opened 2017-10-26
Replace 'httplib' with 'urllib3'

AymericDu:
Replace `httplib` with `urllib3` for :
- `HttpChecker`
- `HttpStat`
- `ContentReaperFilter`
See #1087

### DIFF
--- a/oio/event/filters/content_cleaner.py
+++ b/oio/event/filters/content_cleaner.py
@@ -14,9 +14,7 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 
-from urlparse import urlparse
-from eventlet import Timeout, GreenPile
-from oio.common.http_eventlet import http_connect
+from eventlet import GreenPile
 from oio.event.evob import Event
 from oio.event.consumer import EventTypes
 from oio.event.filters.base import Filter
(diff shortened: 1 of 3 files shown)


oio/event/filters/content_cleaner.py
```diff
@@ -41,19 +41,19 @@ def __init__(self, *args, **kwargs):
                 "ec": self._handle_rawx,
                 "backblaze": self._handle_b2,
         }
+        self.pool_manager = get_pool_manager()
+        self.timeout = urllib3.Timeout(connect=CONNECTION_TIMEOUT,
+                                       read=CHUNK_TIMEOUT)
```
fvennetier: Can you make it possible to override these timeouts in `self.conf`?

oio/event/filters/content_cleaner.py
```diff
@@ -41,19 +41,20 @@ def __init__(self, *args, **kwargs):
                 "ec": self._handle_rawx,
                 "backblaze": self._handle_b2,
         }
+        self.pool_manager = get_pool_manager()
+        self.timeout = urllib3.Timeout(
+            connect=self.conf.get("connection_timeout", CONNECTION_TIMEOUT),
+            read=self.conf.get("chunk_timeout", CHUNK_TIMEOUT))
```
fvennetier: Why didn't you call it `read_timeout`?

PR #1257, opened 2017-11-08
Monitor the beanstalkd server from its producers

### LINKED ISSUE #1194: meta2 should monitor the beanstalkd size of the queue.
Problems are known to happen when the beanstalkd binlog becomes huge.
Though it still accepts messages, it doesn't work properly on the worker wide.
It seems a good idea to periodically monitor the size of the tubes, and stop sending after a configurable limit has been reached.

/cc @racciari

jfsmig:
* Closes #1194 
* Fix memleaks showing up when network errors occur
* Fix retry conditions to avoid looping on event when an unrecoverable error happens
* Enforce the expected consequences to specific replies from beanstalkd
* Log any event dropped (at least its preamble).


events/oio_events_queue_beanstalkd.c
```diff
@@ -151,88 +158,135 @@ _send (int fd, struct iovec *iov, unsigned int iovcount)
...
+		}
+		g_strfreev(lines);
+	}
+
+	const gint64 max_jobs = oio_events_beanstalkd_check_level_deny;
+	if (max_jobs > 0 && total > max_jobs)
+		return BUSY("FULL");
+	const gint64 alert = oio_events_beanstalkd_check_level_alert;
+	if (alert > 0 && total > alert)
+		GRID_WARN("Beanstalkd too loaded");
```
fvennetier: You should print the alert and error limits in the warning (and the error message, since it will also trigger a warning trace).

events/oio_events_queue_beanstalkd.c
```diff
@@ -264,17 +318,62 @@ _q_run (struct oio_events_queue_s *self, gboolean (*running) (gboolean pending))
...
+		/* Maybe reconnect to the beanstalkd */
+		if (fd < 0) {
+			GError *err = NULL;
+			fd = sock_connect(q->endpoint, &err);
+			if (!err)
+				err = _poll_out (fd);
+			if (!err) {
+				err = _use_tube (fd, q->tube);
+				if (intercept_errors)
+					(*intercept_errors) (err);
```
fvennetier: Here you may want to prefix the error to say the TCP connection is OK but the tube selection failed. This could help in case of a TCP port mismatch where the listening service is not a beanstalkd.

PR #1259, opened 2017-11-09
Add support for IPv6

AymericDu:
To test with `oio-reset.sh`:
```shell
$ beanstalkd -l ::1 &
$ export ZK=::1:2181
$ oio-reset.sh -f oio-sds/etc/bootstrap-preset-smallrepli.yml -I ::1
```

### DIFF
--- a/oio/conscience/checker/tcp.py
+++ b/oio/conscience/checker/tcp.py
@@ -24,11 +24,14 @@ class TcpChecker(BaseChecker):
             if k not in self.checker_conf:
                 raise exc.ConfigurationException(
                     'Missing field "%s" in configuration' % k)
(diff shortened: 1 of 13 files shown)


oio/conscience/checker/tcp.py
```diff
@@ -24,11 +24,14 @@ def configure(self):
             if k not in self.checker_conf:
                 raise exc.ConfigurationException(
                     'Missing field "%s" in configuration' % k)
-        self.addr = (self.checker_conf['host'], self.checker_conf['port'])
+        addrinfo = socket.getaddrinfo(
```
jfsmig: Be carefull to pass the option to avoid a costly DNS resolution.
We currently only manage addresses so you can give the AI_NUMERICHOST option.

PR #1267, opened 2017-11-17
Backup/Restore: manage checksums

murlock:
Backup now save chunks checksums in manifest.
Restore ensure that each part match checksums.

### DIFF
--- a/oio/container/backup.py
+++ b/oio/container/backup.py
@@ -712,6 +712,24 @@ class ContainerRestore(object):
         if (self.mode == self.MODE_FULL
                 or self._range[1] == self.cur_state['last_block']):
             code = 201
+
+            manifest = self.cur_state.get('manifest')
+            if manifest:
+                for entry in self.cur_state.get('manifest'):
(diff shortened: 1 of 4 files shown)


oio/container/backup.py
```diff
@@ -506,24 +621,38 @@ def parse_reg_type(self, hdrs, account, container):
...
                 raise BadRequest("Invalid internal state")
+
+            # since an error has occured, we have to reset
+            # current offset to start of current chunk
+            # and remove stored checksum if set
+
+            if isinstance(ex, exc.SourceReadError):
+                self.logger.error("Invalid checksum detected for %s",
+                                  self.inf.name)
+                raise BadRequest("Checksum error %s" % self.inf.name)
```
fvennetier: How do we know this error comes from a bad checksum?

murlock: Hum, I'm agree that it is a shortcut... I will add a flag in LimitedStream and check here to remove ambiguity regarding other possible errors

PR #1269, opened 2017-11-17
Python3 compatibility

dimtion:
This pull request add some Python3 compatibility to the project while maintaining python2 compat.

Most of the work was done tweaking languages incompatibilities between py2&py3, and should not introduce any regression to the Python 2 build.

However two main issues should be doubled checked:
* Some str/byte/unicode errors could be made during the update process. Maybe more tests should be added to ensure that no regression was added here.
* The commit bddda91 introduce some modifications to the test suite, and potentially some regressions to the interface of some functions. Not knowing well the code base, I tried to do my best.

The PR is still a WIP, as of today still 20 tests are not passing on my device, please give me feedback if you want me to continue this work, as we urge to need python3 support at least for the SDK part.

I can also open an issue if you want to discuss.

### DIFF
--- a/all-requirements.txt
+++ b/all-requirements.txt
@@ -11,3 +11,4 @@ requests<2.13.0
 simplejson>=2.0.9
 urllib3>=1.13.1
 werkzeug>=0.9.1
+six==1.11.0
--- a/oio/account/backend.py
+++ b/oio/account/backend.py
@@ -16,6 +16,7 @@
 
 from time import time
 
+from six import text_type
 import redis
 import redis.sentinel
 from werkzeug.exceptions import NotFound, Conflict, BadRequest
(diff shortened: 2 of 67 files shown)


dimtion:
I've merged the master onto this branch, tell me if you prefer a simple rebase.

It seems that the buggy build are because of the master branch.

fvennetier:
Hello. Yes, the build fails because of a test of the pkg-config version. I tried to fix it this morning. I hope it is OK now.

fvennetier:
Hello. Thanks for your work. Indeed, we prefer a rebase, to keep the history clean. I did it here https://github.com/fvennetier/oio-sds/commits/dimtion-python3. If it's OK, you can force-push into your branch to update this pull-request.

dimtion:
Ok, so it seems good now. I also have this patch https://github.com/dimtion/oio-sds/commit/786505b88747f920eb6810246fe8a9226d85067f that I didn't included in the PR. The tests don't fully pass under python3 (that is why I didn't included it), but for end users that only want the OpenIO SDK it seems enough.

PR #1275, opened 2017-11-21
Set a longer+configurable timeout for beanstalkd requests

jfsmig:


events/oio_events_queue_beanstalkd.c
```diff
@@ -138,7 +141,7 @@ _send (int fd, struct iovec *iov, unsigned int iovcount)
 			struct pollfd pfd = {};
 			pfd.fd = fd;
 			pfd.events = POLLOUT;
-			metautils_syscall_poll (&pfd, 1, 1000);
+			metautils_syscall_poll (&pfd, 1, timeout);
```
fvennetier: Out of scope: we retry even if poll returns an error?

PR #1277, opened 2017-11-21
Document some service API calls

jfsmig:


rawx-apache2/src/rawx_chunk_update.c
```diff
@@ -95,6 +95,96 @@ __get_chunkupdate_resource(const request_rec *r, const dav_hooks_repository *hoo
...
+     - Description
+   * - CHUNKID
+     - a string of 64 hexadecimal characters
+
+
+.. list-table:: Headers
+   :header-rows: 1
+   :widths: 10 20
+
+   * - Header
```
murlock: Minor: we should show which headers are mandatory or optional

PR #1290, opened 2017-11-27
Improve integrity crawler

fvennetier:


oio/crawler/integrity.py
```diff
@@ -28,9 +28,11 @@
 from eventlet.greenpool import GreenPool
 
 from oio.common import exceptions as exc
+from oio.common.storage_method import STORAGE_METHODS
 from oio.account.client import AccountClient
 from oio.container.client import ContainerClient
 from oio.blob.client import BlobClient
+from oio.api.object_storage import _sort_chunks
```
jfsmig: It is sad to export a `_` prefixed function

murlock: I've also imported this function in `backup.py`

fvennetier: I can rename it, but not in 4.1.x branch. In 4.x branch it has been moved, but still has the leading underscore. https://github.com/open-io/oio-sds/blob/4.x/oio/common/storage_functions.py#L113.

PR #1310, opened 2017-12-14
C API: Implement the rollback of uploaded chunks

jfsmig:
Multiplexed libcurl calls though the curl_multi API.

### DIFF
--- a/core/sds.c
+++ b/core/sds.c
@@ -29,10 +29,10 @@ License along with this library.
 #include <curl/curlver.h>
 
 #include <core/client_variables.h>
-#include <core/oiostr.h>
 #include <metautils/lib/metautils.h>
 
 #include "http_put.h"
+#include "http_del.h"
 #include "http_internals.h"
 #include "internals.h"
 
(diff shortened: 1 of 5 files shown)


core/sds.c
```diff
@@ -1828,15 +1854,17 @@ _upload_sequential (struct oio_sds_s *sds, struct oio_sds_ul_dst_s *dst,
...
-		if (e) {
-			GRID_WARN("Upload abort failed: (%d) %s",
-					oio_error_code (e), oio_error_message (e));
-			oio_error_free (e);
+		if (oio_error_code(err) == CODE_CONTENT_EXISTS ||
+				oio_error_code(err) == CODE_CONTENT_PRECONDITION) {
+			_upload_abort_no_error(ul);
+		} else {
+			GRID_WARN("Conditons unsafe to abort the upload: (%d) %s",
+					oio_error_code(err), oio_error_message(err));
```
fvennetier: We should not print a warning when the commit step succeeded.

core/sds.c
```diff
@@ -1828,15 +1854,17 @@ _upload_sequential (struct oio_sds_s *sds, struct oio_sds_ul_dst_s *dst,
...
 		err = oio_sds_upload_commit (ul);
-	else {
-		struct oio_error_s *e = oio_sds_upload_abort (ul);
-		if (e) {
-			GRID_WARN("Upload abort failed: (%d) %s",
-					oio_error_code (e), oio_error_message (e));
-			oio_error_free (e);
+		if (oio_error_code(err) == CODE_CONTENT_EXISTS ||
+				oio_error_code(err) == CODE_CONTENT_PRECONDITION) {
+			_upload_abort_no_error(ul);
```
fvennetier: Since oio-sds 4.1.9, both `CODE_CONTENT_EXISTS` and `CODE_CONTENT_PRECONDITION` are converted to `HTTP_CODE_CONFLICT` by the proxy, thus we must only check for this one.

jfsmig: Sadly, the C proxy client (at core/proxy.c +395) doesn't really take care of the HTTP response code and always tries to parse the body as a JSON error (object with "status" and "message").

PR #1311, opened 2017-12-18
Fix service stats, add "stat.event.health"

fvennetier:


events/oio_events_queue_beanstalkd.c
```diff
@@ -478,3 +484,19 @@ _q_is_stalled (struct oio_events_queue_s *self)
...
+static gint64
+_q_get_health(struct oio_events_queue_s *self)
+{
+	struct _queue_BEANSTALKD_s *q = (struct _queue_BEANSTALKD_s*) self;
+	EXTRA_ASSERT(q != NULL && q->vtable == &vtable_BEANSTALKD);
+
+	gint64 limit = G_MAXINT64;
+	if (oio_events_beanstalkd_check_level_deny > 0)
+		limit = oio_events_beanstalkd_check_level_deny;
+	else if (oio_events_beanstalkd_check_level_alert > 0)
```
jfsmig: This is incorrect. That configuration variable can only be used for warnings to be displayed and must not trigger the zeroing of a service's score.

events/oio_events_queue_beanstalkd.c
```diff
@@ -478,3 +484,19 @@ _q_is_stalled (struct oio_events_queue_s *self)
...
+		limit = oio_events_beanstalkd_check_level_alert;
+
+	// TODO: take g_async_queue_length (q->queue) into account
+	return 100 - (q->pending_events * 100 / limit);
```
jfsmig: I propose to reserve this piece of logic for the case of an explicitly configured hard limit with `oio_events_beanstalkd_check_level_deny`.

But when no configuration has been set, the limit at G_MAXINT64 will ~always give a 100% health and be unusable to compare the service using their score. We could use a function smoothly decreasing, e.g. `1 + 1/ln(x+1)`.

events/oio_events_queue_beanstalkd.c
fvennetier: This does not trigger the zeroing of the service.

PR #1312, opened 2017-12-18
Improve meta2 content events

fvennetier:
- do not send chunks in "storage.content.new" events
- add "part" and "parts" entries in "storage.content.deleted" events,
  representing respectively the id of the current part and the total
  number of parts of the event.


fvennetier:
Which job are you talking about?
I read somewhere that someone was trying to distinguish broken events from split events in which there is no alias part. This patch:
* Prevents sending `storage.content.new` without an alias part.
* Adds part numbers to `storage.content.deleted` so you can distinguish the first (with alias part) from the others (without alias part).

Issue #903, opened 2017-03-03
Directory bootstrap failure on lowend servers


racciari:
When bootstrapping a lowend server like a RaspberryPi, it ends with a HTTP500 failure:

```
root@raspberrypi:/home/pi# openio --oio-ns=OPENIO directory bootstrap --no-rdir -vvv --debug
admin API: cmd group openio.admin
directory API: cmd group openio.directory
storage API: cmd group openio.storage
command: directory bootstrap -> oio.cli.admin.directory.DirectoryInit
take_action(Namespace(rdir=False, replicas=3))
Using parameters {'admin_mode': False, 'log_level': 'DEBUG', 'is_cli': True, 'namespace': 'OPENIO', 'proxyd_url': 'http://192.168.1.49:6006', 'account_name': ''}
Starting new HTTP connection (1): 192.168.1.49
"GET /v3.0/OPENIO/conscience/list?type=meta1 HTTP/1.1" 200 132
Computing meta1 prefix mapping...
0%
10%
20%
30%
40%
50%
60%
70%
80%
90%
100%
Equilibrating...
No equilibration possible when meta1_digits is set to 0
Saving...
Starting new HTTP connection (1): 192.168.1.49
"POST /v3.0/OPENIO/admin/meta0_force HTTP/1.1" 500 32
Timeout (HTTP 500) (STATUS 7)
Traceback (most recent call last):
  File "/usr/lib/python2.7/dist-packages/cliff/app.py", line 374, in run_subcommand
    result = cmd.run(parsed_args)
  File "/usr/lib/python2.7/dist-packages/cliff/command.py", line 54, in run
    self.take_action(parsed_args)
  File "/usr/lib/python2.7/dist-packages/oio/cli/admin/directory.py", line 56, in take_action
    mapping.force()
  File "/usr/lib/python2.7/dist-packages/oio/directory/meta0.py", line 126, in force
    self.m0.force(self.to_json().strip())
  File "/usr/lib/python2.7/dist-packages/oio/directory/meta0.py", line 26, in force
    self._request('POST', uri, data=mapping)
  File "/usr/lib/python2.7/dist-packages/oio/common/client.py", line 51, in _request
    return self._direct_request(method, url, **kwargs)
  File "/usr/lib/python2.7/dist-packages/oio/common/client.py", line 45, in _direct_request
    raise from_response(resp, body)
ClientException: Timeout (HTTP 500) (STATUS 7)
[...]

PR #1314, opened 2017-12-22
Python API: add support for operation deadlines

fvennetier:


oio/api/base.py
```diff
@@ -104,6 +108,17 @@ def _direct_request(self, method, url, headers=None, data=None, json=None,
         if self.admin_mode or admin_mode:
             out_headers[ADMIN_HEADER] = '1'
 
+        # Look for a request deadline, deduce the timeout from it.
```
jfsmig: Is it possible to export this piece of logic into a method responsible to patch the kwargs?
The `_direct_request()` method is already really long.

fvennetier: I'll think about it.

jfsmig:
Will help solving #903

Issue #1291, opened 2017-11-28
No event managed when a reference is created.


jfsmig:
When I call `openio --oio-account=ACCT reference create MYREF` and it is a success, I expect the command `openio account list` to display `ACCT` soon after.

The chain between the directory of services (meta1) and the account service is broken somewhere.

jfsmig:
**managed** has been chosen especially for "you know who"

PR #1315, opened 2017-12-29
Send event when a reference is created

AymericDu:
See #1291

### DIFF
--- a/meta1v2/meta1_backend_references.c
+++ b/meta1v2/meta1_backend_references.c
@@ -142,6 +142,9 @@ meta1_backend_user_create(struct meta1_backend_s *m1,
 			}
 		}
 		err = sqlx_transaction_end(repctx, err);
+
+		if (!err)
+			meta1_backend_notify_services(m1, url);
 	}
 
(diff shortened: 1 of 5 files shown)


meta1v2/meta1_backend_references.c
```diff
@@ -110,6 +110,17 @@ __destroy_container(struct sqlx_sqlite3_s *sq3, struct oio_url_s *url,
...
+static void
+__notify_account(struct meta1_backend_s *m1, struct oio_url_s *url)
+{
+	if (!m1->notifier)
+		return;
+
+	GString *notif = oio_event__create("storage.account.new", url);
+	g_string_append_static (notif, ",\"data\":[]}");
+	oio_events_queue__send(m1->notifier, g_string_free (notif, FALSE));
+}
```
fvennetier: I think we would better reuse the `account.services` event type, and send an empty list of services. Unfortunately we never managed these events, but they are still send when the list of services linked to a reference changes.
The code is in https://github.com/open-io/oio-sds/blob/master/meta1v2/meta1_backend_services.c#L1217, you will have to refactor a little bit.

Issue #918, opened 2017-03-17
Implement container flush


GuillaumeDelaporte:
It could be very interesting to have something like `openio container flush mycontainer` or `openio object delete mycontainer *`

gdelafond:
It could be very dangerous, I think we need to have a confirmation before doing the job. Maybe with a "bypass" to be able to automate it with an env variable, or a --force option ?
```
$ openio container flush mycontainer
X objects will be deleted from "mycontainer", are you sure ? (y/N)
```

With env variable: 
`$ ALLOW_CONTAINER_FLUSH=1 openio container flush mycontainer`

With --force option: 
`$ openio container flush --force mycontainer`

Any thoughts about this ?

fvennetier:
The delete with wildcard is dangerous (you want to `delete a*` but you type `delete *`). When you type `flush` there is no possible confusion. I think asking for a confirmation is excessive.

aneutron:
Can be very useful when a restore operation goes wrong and we need to delete the container.

PR #1316, opened 2018-01-02
Flush container

AymericDu:
See #918


jfsmig:
Please, can you rebase your commits on top of 4.x ?

meta2v2/meta2_utils.c
```diff
@@ -2738,9 +2738,41 @@ m2db_deduplicate_contents(struct sqlx_sqlite3_s *sq3, struct oio_url_s *url)
...
+	GVariant *params[3] = {NULL};
+	gchar sql[16];
+	g_snprintf(sql, 16, "1 LIMIT %"G_GINT64_FORMAT, limit);
+	err = ALIASES_load(sq3->db, sql, params, _bean_buffer_cb, aliases);
+	metautils_gvariant_unrefv(params);
+
+	guint nb_aliases = aliases->len;
+	if (nb_aliases == limit) {
+		g_ptr_array_remove_index_fast(aliases, limit-1);
+		*truncated = TRUE;
```
fvennetier: The removed alias should be freed somehow.

tests/functional/api/test_objectstorage.py
```diff
@@ -275,6 +275,72 @@ def test_container_set_properties(self):
...
+        flush_and_check()
+
+        # many contents
+        for i in range(1024):
+            self.api.object_create(self.account, name,
+                                   obj_name='content'+str(i), data='data')
+        flush_and_check()
+
+        # give meta2 and beanstalk time to catch their breath
+        time.sleep(30)
```
fvennetier: Couldn't we wait for the scores to raise a little instead of arbitrary sleeping this huge amount of time?

oio/cli/container/container.py
```diff
@@ -215,6 +215,27 @@ def take_action(self, parsed_args):
...
+            metavar='<container>',
+            help='Container to flush'
+        )
+        return parser
+
+    def take_action(self, parsed_args):
+        self.log.debug('take_action(%s)', parsed_args)
+
+        self.app.client_manager.storage.container_flush(
+            self.app.client_manager.account, parsed_args.container)
```
fvennetier: Is it wanted to use the "slow" version?

AymericDu: In a first phase, we don't give access to the fast version from the CLI. Because the fast version is really brutal for meta2, beanstalk and account.

meta2v2/meta2_backend.c
```diff
@@ -720,11 +714,24 @@ meta2_backend_flush_container(struct meta2_backend_s *m2, struct oio_url_s *url)
...
+				int rc = sqlx_exec(sq3->db, "VACUUM");
+				if (rc != SQLITE_OK)
+					err = SQLITE_GERROR(sq3->db, rc);
+			}
+		}
+		if (!err) {
+			if (!(err = sqlx_transaction_begin(sq3, &repctx))) {
+				if (!err)
+					sqlx_transaction_notify_huge_changes(repctx);
+				err = sqlx_transaction_end(repctx, err);
```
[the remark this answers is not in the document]
AymericDu: Before, I execute a `VACUUM`. And the `VACUUM` can not be executed in a transaction. So I close the first transaction, then open a second transaction to indicate the change.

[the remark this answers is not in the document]
fvennetier: OK I got it.
I think the `sqlx_transaction_notify_huge_changes` should be executed only at the end of the flush (when the result is no more truncated). Its purpose is to trigger a full copy of the database, which is only interesting when the file is small (i.e. after the `VACUUM`).

PR #1317, opened 2018-01-03
Disallow using an unknown storage policy

AymericDu:
See #1153

### DIFF
--- a/proxy/m2_actions.c
+++ b/proxy/m2_actions.c
@@ -1205,8 +1205,9 @@ _m2_container_create (struct req_args_s *args, struct json_object *jbody)
 	if (err)
 		return _reply_m2_error(args, err);
 
-	err = _m2_container_create_with_properties(
-			args, properties, OPT("stgpol"), OPT("verpol"));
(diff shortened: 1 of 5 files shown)


proxy/m2_actions.c
```diff
@@ -1192,6 +1192,16 @@ action_m2_container_propdel (struct req_args_s *args, struct json_object *jargs)
 	return action_sqlx_propdel(args, jargs);
 }
 
+static char *
+_get_property(char **properties, char *key)
```
jfsmig: This function has a possible general interest, please move it in `core/str.c` and export it with a name consistent with the remaining of the file (maybe `KV_get_key`).

proxy/m2_actions.c
```diff
@@ -1192,6 +1192,16 @@ action_m2_container_propdel (struct req_args_s *args, struct json_object *jargs)
...
+{
+	for (gchar **p=properties; *p && *(p+1) ;p+=2) {
```
jfsmig: If generalizing this function, a check on the non-nullity of `properties` is necessary.

PR #1318, opened 2018-01-05
Temporary disable new plyvel versions, they don't build properly for us

vincent-legoll:
You can reproduce by just cloning openio/oio-qa and building the docker image

The (fairly new plyvel 1.0.0 from 2 days ago) broke it, so stick with the working one until further investigated & fixed...

Error was:
    plyvel/_plyvel.cpp: In function 'int __pyx_f_6plyvel_7_plyvel_parse_options(leveldb::Options*, bool, bool, PyObject*, PyObject*, PyObject*, PyObject*, PyObject*, PyObject*, PyObject*, PyObject*, int, PyObject*, PyObject*)':
    plyvel/_plyvel.cpp:3341:22: error: 'struct leveldb::Options' has no member named 'max_file_size'
         __pyx_v_options->max_file_size = __pyx_t_4;

### DIFF
--- a/Variables.md
+++ b/Variables.md
@@ -85,6 +85,22 @@ Used by `gcc`
 
 ### Variables for production purposes
 
+### client.down_cache.avoid
+
+> Should an error be raised when the peer is marked down, instead of trying to contact the peer.
+
+ * default: **FALSE**
+ * type: gboolean
+ * cmake directive: *OIO_CLIENT_DOWN_CACHE_AVOID*
+
+### client.down_cache.shorten
+
+> Should the connection timeout be dramatically shortened when talking to a peer that has been reported down. Set to false by default, this is evaluated after the avoidance of those peers.
+
+ * default: **FALSE**
+ * type: gboolean
+ * cmake directive: *OIO_CLIENT_DOWN_CACHE_SHORTEN*
+
 ### client.errors_cache.enabled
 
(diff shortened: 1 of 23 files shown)


vincent-legoll:
You can reproduce by just cloning openio/oio-qa and building the docker image

all-requirements.txt
```diff
@@ -2,7 +2,7 @@ cliff>=2.0.0,<2.1.0
 eventlet!=0.18.3,!=0.20.1,<0.21.0,>=0.18.2
 gunicorn>=19.4.5
 lxml
-plyvel>=0.9
+plyvel==0.9
```
fvennetier: If we know plyvel 1.0.0 is broken, can't we just exclude this version? I'm thinking of something like `plyvel>=0.9,!=1.0.0`.

vincent-legoll:
The (fairly new 1.0.0 from 2 days ago) broke it, so stick with the working one until further investigated & fixed...

all-requirements.txt
vincent-legoll: This is only a temporary workaround, likely to be reverted when a proper fix is found...

fvennetier: plyvel 1.0.0 requires leveldb 1.20, and our Travis build installs `libleveldb1` 1.15.0 (and there is no superior version in this distribution), thus we are stuck with `plyvel<1.0.0`. Now I'm suggesting we use `plyvel>=0.9,<1.0.0` in case someone creates updates over the 0.9 version.

PR #1321, opened 2018-01-10
oio-crawler-integrity: Propose a lightweight presence check

jfsmig:
We spotted a recent raise of the interest in a simple & lightweight presence check for contents. The `rawx` services loaded all the xattr of each chunk, inducing an unexpectedly high load on the devices. The purpose of the current PR is to propose a backward-compliant lightweight check that just validates the presence (but not the consistency) of each content, thus avoiding to touch the xattr/payload of each chunk.


rawx-apache2/src/rawx_repository.c
```diff
@@ -189,18 +189,18 @@ dav_rawx_get_resource(request_rec *r, const char *root_dir, const char *label,
...
-				 RESOURCE_STAT_CHUNK_READ_ATTRS : 0;
-	if (r->method_number == M_PUT || r->method_number == M_POST)
+	int flags = 0;
+
+	if (r->method_number == M_GET) {
+		const char *value = apr_table_get(r->headers_in, "X-oio-xattr");
+		if (oio_str_parse_bool(value, TRUE))
+			flags |= RESOURCE_STAT_CHUNK_READ_ATTRS;
+	} else if (r->method_number == M_OPTIONS) {
+		flags |= RESOURCE_STAT_CHUNK_READ_ATTRS;
```
fvennetier: Extended attributes are also used to complete the event during DELETE operations. Some of them are useless, but at least `container_id` and `content_id` (along with `chunk_id` and `volume_id`) are used to deindex the chunk from rdir.

jfsmig: :man_facepalming: 

OK, got it.

PR #1322, opened 2018-01-10
[WIP] Use UUID instead of IP:PORT

murlock:
- [x] Include UUID to each service (need more specification for real deployment)
- [x] expose to openio cli
- [ ] add fallback for old service (generate it from UUID)
- [ ] add migration step ? (to be discussed)
- [ ] store UUID instead of IP:PORT in directories
- [ ] add documention


murlock:
Superseded by #1350

Issue #1273, opened 2017-11-21
set_properties on container does not increase admin version


jkasarherou:
It seems a set_properties will not correctly increase admin table version number!!

jfsmig:
That value is purely for internal use, so since the change in the admin table are replicated, is it a problem to have that version [of the admin table] not incremented ?

jfsmig:
Fixed in #1323

PR #1323, opened 2018-01-12
Increase admin version when change properties on container

AymericDu:
See #1273


tests/functional/container/test_container.py
```diff
@@ -320,6 +320,52 @@ def test_create_with_unknown_storage_policy(self):
...
+            self.assertIsInstance(body, dict)
+            self.assertIsInstance(body.get('properties'), dict)
+            self.assertDictEqual(expected, body['properties'])
+            self.assertEqual(expected_version,
+                             body['system']['sys.m2.version'])
+
+        def del_ok(keys):
+            resp = self.request('POST', self.url_container('del_properties'),
+                                params=params, data=json.dumps(keys))
+            self.assertEqual(resp.status, 200)
```
fvennetier: off topic: it's unfortunate that this request returns 200 and an empty JSON object instead of 204 and no body.

tests/functional/container/test_container.py
```diff
@@ -320,6 +320,52 @@ def test_create_with_unknown_storage_policy(self):
...
+        p1 = {random_content(): random_content()}
+
+        get_ok({}, '0')
+        set_ok(p0)
+        get_ok(p0, '1')
+        set_ok(p1)
+        get_ok(merge(p0, p1), '2')
+        del_ok(p0.keys())
+        get_ok(p1, '3')
+        del_ok(p0.keys())
```
fvennetier: Deleting twice the same key increases the version number?

PR #1330, opened 2018-01-17
proxy: Fix the handling of URL of directory services

jfsmig:
* Introduce a test to detect a possible regression
* Avoid an ugly assertion to be raised
* Apply the master/slave preference when contacting meta0 and meta1


proxy/common.c
```diff
@@ -258,12 +258,12 @@ gridd_request_replicated (struct client_ctx_s *ctx, request_packer_f pack)
...
 			return NEWERROR (CODE_CONTAINER_NOTFOUND, "No service located");
-		} else {
-			/* We found some locations, let's keep only the URL part */
-			meta1_urlv_shift_addr (m1uv);
-			/* let's prefer the services requested (master, slave, etc) */
-			_sort_services (ctx, election_key, m1uv);
 		}
+
+		/* We found some locations, let's keep only the URL part */
+		meta1_urlv_shift_addr (m1uv);
```
fvennetier: When merging to 4.x branch, be careful not to shift twice.

tests/functional/directory/test_meta1.py
```diff
@@ -0,0 +1,27 @@
...
+
+class TestProxyFailure(BaseTestCase):
+    def setUp(self):
+        super(TestProxyFailure, self).setUp()
+
+    def test_admin_debug_on_meta1(self):
+        params = {'ref': random_str(64),
+                  'acct': random_str(64),
+                  'type': 'meta1'}
+        self.request('POST', self._url('admin/debug'), params=params)
```
fvennetier: What about meta0?

PR #1340, opened 2018-01-26
Set status field during meta2 database creation

murlock:
When creating a container, status was left unset.

It was only updated when creating a snapshot

### DIFF
--- a/tests/functional/container/test_container.py
+++ b/tests/functional/container/test_container.py
@@ -20,6 +20,8 @@ import logging
 import simplejson as json
 import struct
 from tests.utils import BaseTestCase, random_str
+from oio.common.constants import OIO_DB_STATUS_NAME, OIO_DB_ENABLED, \
+                                 OIO_DB_FROZEN, OIO_DB_DISABLED
 
 
 def random_content():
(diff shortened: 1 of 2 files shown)


tests/functional/container/test_container.py
```diff
@@ -319,6 +321,46 @@ def test_create_with_unknown_storage_policy(self):
         data = self.json_loads(resp.data)
         self.assertEqual(data["status"], 480)
 
+    def _test_create_with_status(self, status=None):
+        def _status(_data):
+            return data['system']['sys.status']
```
fvennetier: I think you mean `_data['system']['sys.status']` (with the undescore). This unexpectedly works because of Python's variable scope.

tests/functional/container/test_container.py
```diff
@@ -319,6 +321,46 @@ def test_create_with_unknown_storage_policy(self):
...
+
+        params = self.param_ref(self.ref)
+        headers = {}
+        headers['x-oio-action-mode'] = 'autocreate'
+        headers['Content-Type'] = 'application/json'
+        params = self.param_ref(self.ref)
```
fvennetier: `params` is assigned twice.

PR #1341, opened 2018-01-26
Apply all python requirements at once

AymericDu:


all-requirements.txt
```diff
@@ -1,7 +1,7 @@
 cliff>=2.0.0,<2.1.0
 # This can be simplified when https://github.com/eventlet/eventlet/issues/401 is fixed
 eventlet!=0.18.3,!=0.20.1,<0.21.0,>=0.18.2
-requests<2.13.0
```
vincent-legoll: You should explain that in the commit msg as it is not obvious why you changed it that way

AymericDu: In the past, we think we had a similar problem: https://github.com/requests/requests/issues/3992
But the experimentation shows that the recent versions work.

PR #1350, opened 2018-02-02
ServiceID for RawX

murlock:
This PR introduce Service ID for RAWX: 
it allows them to update their IP with few command (a `Chunk ID` will use `Service Id` instead `addr`)

To support this, changes were made on:
**proxy**:
- NS/conscience/resolve/#GET: retrieve `addr:ip` from an `uuid` (used by `oio/blob`)
- /cache/show/#GET : retrieve current cache of proxy 
- `content/prepare`: as `url` field contains ID of chunks, not id, it adds a field `real_url` with resolved Service Id 
**core/lb**:
- struct is enriched with `addr` field and try to use Service Id instead `addr`
- new function to iterate over items in lb_world_s 
**oio.blob**:
- add cache system as event contains Service Id, each entry can be used 60 seconds.
**oio**:
- use new url if set
**cli**
- to reflect changes, `openio cluster (localt) list` will show `Addr` and `ServiceId` instead of ambiguous `Id`


oio/blob/cache.py
```diff
@@ -0,0 +1,57 @@
...
+        self.pool_manager = pool_manager or get_pool_manager()
+        self._client = ProxyClient(self.conf, pool_manager=self.pool_manager,
+                                   request_prefix="/conscience")
+        self.logger = get_logger(conf)
+
+    def _get_addr(self, item):
+        if item in self._cache and self._cache[item]['ts'] < time():
+            return self._cache[item]['addr']
+        resp, body = self._client._request(
+            "GET", "/resolve", params={'type': 'rawx', 'uuid': item})
```
fvennetier: The host resolving request (without caching) could be implemented in `ConscienceClient`.

oio/cli/cluster/cluster.py
```diff
@@ -76,27 +76,28 @@ def _list_services(self, parsed_args):
...
                               slots, up, score)
                 yield values
 
     def take_action(self, parsed_args):
         self.log.debug('take_action(%s)', parsed_args)
         if parsed_args.stats:
-            columns = ('Type', 'Id', 'Volume', 'Location', 'Slots', 'Up',
-                       'Score', 'Stats')
+            columns = ('Type', 'Id', 'Uuid', 'Volume', 'Location', 'Slots',
+                       'Up', 'Score', 'Stats')
```
fvennetier: We should keep the `Id` column for service IDs, and add an `Addr` (or `NetLoc`?) column for resolved network addresses.

oio/event/filters/content_cleaner.py
```diff
@@ -35,23 +35,24 @@ def __init__(self, *args, **kwargs):
                 "ec": self._handle_rawx,
                 "backblaze": self._handle_b2,
         }
-        self.blob_client = BlobClient()
+        self.blob_client = BlobClient(args[1])
```
fvennetier: We should use `self.conf` (the `Filter` superclass saves it).

proxy/cs_actions.c
```diff
@@ -135,6 +135,55 @@ conscience_remote_remove_services(gchar **allcs, const char *type, GSList *ls, g
...
+conscience_resolve_uuid(gchar **cs UNUSED, const char *type, const char *uuid, gchar **out) {
+    *out = NULL;
+
+    gchar *key = oio_make_service_key(ns_name, type, uuid);
+    struct oio_lb_item_s *item = oio_lb_world__get_item(lb_world, key);
+    g_free(key);
+    if (item) {
+        *out = g_strdup(item->addr);
+        return NULL;
+    }
```
fvennetier: https://www.emacswiki.org/pics/static/TabsSpacesBoth.png

PR #1358, opened 2018-02-12
Python API: expose backoff_factor option

murlock:
By default, urllib3 doesn't use backoff when retries are
enabled. This commit introduce parameter to update it.

Following urllib3 documentation, backoff is used after second
try and will sleep by using formula:
{backoff factor} * (2 ^ ({number of total retries} - 1))

### DIFF
--- a/oio/common/http.py
+++ b/oio/common/http.py
@@ -31,6 +31,7 @@ READ_TIMEOUT = 30.0
 
 DEFAULT_POOLSIZE = 32
 DEFAULT_RETRIES = 0
+DEFAULT_BACKOFF = 0
 
 
 class CustomHTTPResponse(HTTPResponse):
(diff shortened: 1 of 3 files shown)


oio/common/http.py
```diff
@@ -268,11 +270,14 @@ def get_pool_manager(pool_connections=DEFAULT_POOLSIZE,
     :type pool_maxsize: `int`
     :param max_retries: number of retries per request
     :type max_retries: `int`
+    :param backoff_factor: backoff factor to apply between after second try
```
fvennetier: *to apply between **attempts** after second try*

oio/common/http.py
```diff
@@ -258,7 +259,8 @@ def pop(self, k, default=None):
 
 def get_pool_manager(pool_connections=DEFAULT_POOLSIZE,
                      pool_maxsize=DEFAULT_POOLSIZE,
-                     max_retries=DEFAULT_RETRIES):
+                     max_retries=DEFAULT_RETRIES,
+                     backoff_factor=DEFAULT_BACKOFF):
```
fvennetier: Don't forget to add `'backoff_factor'` [here](https://github.com/open-io/oio-sds/blob/4.1.12/oio/api/base.py#L26).

PR #1361, opened 2018-02-14
Manage multiple beanstalkd

jfsmig:


oio/event/beanstalk.py
```diff
@@ -413,18 +415,25 @@ class Beanstalk(object):
 
     @classmethod
     def from_url(cls, url, **kwargs):
+        if url is None or not url:
```
fvennetier: `if not url` is sufficient.

Issue #1163, opened 2017-09-07
Versioning is always unlimited regardless the configuration of the configuration


GuillaumeDelaporte:
Trying to set the max number of versions for an object, in the namespace configuration or forcing it at the container level, the number of versions is always unlimited.

fvennetier:
After PR #1250, versioning is still not enforced, but you can purge exceeding versions by running `openio container purge`. I see 3 possible evolutions:
1. refuse to add a new version when the limit is reached;
2. purge oldest versions on-the-fly (this is the original behavior);
3. keep the code as-is, and let the *Lifecycle* crawler purge the oldest versions.

PR #1363, opened 2018-02-15
CLI, lifecycle: delete exceeding object versions

AymericDu:
See #1163


oio/api/object_storage.py
```diff
@@ -1182,6 +1182,25 @@ def _m2_chunk_bean(url, meta, content):
                 "pos": meta["pos"],
                 "content": content}
 
+    def object_head(self, account, container, obj, full=False, **kwargs):
+        """
+        Check for the presence of an object in a container.
+
+        :param full: if true, check the presence of the chunks in the rawx
+        :type full: `bool`
+        """
```
fvennetier: I propose we extend this API with a `depth` parameter instead of `full`.
- `depth=0`: do not check chunks;
- `depth=1`: check if there are enough chunks to read the object;
- `depth=2`: check if all chunks are present.

tests/functional/container/test_lifecycle.py
```diff
@@ -402,3 +402,59 @@ def test_execute_expiration_on_missing_objects(self):
...
+    def test_execute_exceeding_version_expiration_without_versioning(self):
+        source = """
+        <LifecycleConfiguration>
+            <Rule>
+                <ID>rule1</ID>
+                <Filter>
+                </Filter>
+                <Status>Enabled</Status>
+                <ExceedingVersionExpiration>
+                </ExceedingVersionExpiration>
```
fvennetier: To be consistent, we may need to put that inside `<NonCurrentVersionExpiration>`, and make it configurable (independently from the versioning configuration of the container).

oio/api/object_storage.py
fvennetier: In the scope of this pull-request, you can stop at `trust_level=0` and `trust_level=2`, letting `trust_level=1` to another pull-request.

oio/cli/container/container.py
```diff
@@ -533,14 +533,25 @@ def get_parser(self, prog_name):
...
+        parser.add_argument(
+            '--max-versions', '--versioning',
+            metavar='<n>',
+            type=int,
+            help="""The number of versions to keep
+ (override the container configuration).
+ n<0 is unlimited number of versions.
+ n=0 is 1 version.
+ n>0 is n versions.
+"""
```
fvennetier: Specifying max-versions < 1 is not relevant.

PR #1375, opened 2018-03-02
Improve WORM mode

fvennetier:
WORM mode used to prevent DELETE operations. Now it also prevents PUT operations that could overwrite existing objects.


meta2v2/meta2_backend.c
```diff
@@ -1014,6 +1019,7 @@ meta2_backend_force_alias(struct meta2_backend_s *m2b, struct oio_url_s *url,
 		args.sq3 = sq3;
 		args.url = url;
 		args.ns_max_versions = meta2_max_versions;
+		args.worm_mode = oio_ns_mode_worm && !oio_ext_is_admin();
```
AymericDu: In `m2db_force_alias`, you don't check if worm mode is enabled.

PR #1382, opened 2018-03-13
Increase the timeouts involved in 'openio volume bootstrap'

jfsmig:
We now accept to lose 30s while connecting and 90s to the overall result.


oio/common/http.py
```diff
@@ -26,8 +26,8 @@
 
 urllib3 = patcher.import_patched('urllib3.__init__')
 
-CONNECTION_TIMEOUT = 2.0
-READ_TIMEOUT = 30.0
+CONNECTION_TIMEOUT = 10.0
```
murlock: CONNECTION_TIMEOUT is also exposed by oio/io.py, but http.CONNECTION_TIMEOUT seems only used by oio/base.py  (but they have now have different values).

oio/cli/admin/volume.py
```diff
@@ -219,13 +219,15 @@ def take_action(self, parsed_args):
 
         try:
             all_rawx = self.app.client_manager.admin.rdir_lb.assign_all_rawx(
-                    parsed_args.max_per_rdir)
+                    parsed_args.max_per_rdir,
+                    connection_timeout=30.0, read_timeout=90.0)
         except ClientException as exc:
             if exc.status != 481:
```
murlock: If exception is due to timeout, we'll hit the raise ?
maybe add a new except OioTimeout ?

PR #1385, opened 2018-03-15
meta1: Avoid losing the cache of managed prefixes

jfsmig:
Add checks around the refreshes of a meta0 cache in the meta1 services.


meta1v2/meta1_prefixes.c
```diff
@@ -322,6 +372,20 @@ meta1_prefixes_get_peers(struct meta1_prefixes_set_s *m1ps,
 	return a;
 }
 
+gboolean
+meta1_prefixes_is_empty(struct meta1_prefixes_set_s *m1ps)
```
AymericDu: This function isn't used

meta1v2/meta1_prefixes.c
```diff
@@ -78,6 +100,22 @@ meta1_prefixes_manage_all(struct meta1_prefixes_set_s *m1ps)
 
 /* NS operations ------------------------------------------------------------ */
 
+static guint8*
+_cache_from_all_m0l(const GSList *l)
+{
+	guint8 *result = g_malloc0(8192);
```
kamel-rahim: Where does `8192` come from?

jfsmig: The meta0 size is 64ki bits, 8ki bytes.
8192 is the size of the bitmap for managed prefixes.

PR #1395, opened 2018-03-20
meta*: Deny requests upon I/O errors

jfsmig:
* oio-conscience-agent: Consider errors from /stat requests a enough to mark the service down
* server/transpport_gridd.c: Keep track of IO errors/success timestamps, if the last activity is an error or if the timestamp of the last success is not recent (stalled), consider the volume is faulty.
* server/transpport_gridd.c: Deny alll but specific local requests with a code 503 that won't fail the general call, and allow the caller to retry on peers.
* sqlx/sqlx_service.c: Implement a probe in a background thread that periodically checks for directory and file creations.

known limitation: the support of the REQ_PING is really limited in the conscience-agent, the return code is not even checked. We could save the REQ_STAT request if we immediately consider the service is down. then we could even allow the REQ_STAT to return complete information.

### DIFF
--- a/oio/conscience/stats/http.py
+++ b/oio/conscience/stats/http.py
@@ -71,8 +71,6 @@ class HttpStat(BaseStat):
                 result = self._parse_func(resp.read())
             else:
                 raise Exception("status code != 200: %s" % resp.status)
-        except Exception as e:
-            self.logger.debug("get_stats error: %s", e)
         finally:
             if resp:
                 try:
--- a/oio/conscience/stats/http.py
+++ b/oio/conscience/stats/http.py
@@ -71,10 +71,10 @@ class HttpStat(BaseStat):
                 result = self._parse_func(resp.read())
             else:
                 raise Exception("status code != 200: %s" % resp.status)
+            return result
         finally:
             if resp:
                 try:
                     resp.force_close()
                 except Exception:
                     pass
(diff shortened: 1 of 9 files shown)


oio/conscience/stats/http.py
```diff
@@ -71,8 +71,6 @@ def get_stats(self):
                 result = self._parse_func(resp.read())
             else:
                 raise Exception("status code != 200: %s" % resp.status)
-        except Exception as e:
-            self.logger.debug("get_stats error: %s", e)
```
fvennetier: The return statement in the `finally` block will prevent any exception from being raised from this method. It should be moved here.

sqlx/sqlx_service.c
```diff
@@ -1034,6 +1036,54 @@ _worker_clients(gpointer p)
...
+	rc = g_file_set_contents(path, "", 0, &err);
+	errsav = errno;
+	(void) g_unlink(path);
+	if (!rc) {
+		GRID_WARN("I/O error on %s: (%d) %s", path, err->code, err->message);
+		g_clear_error(&err);
+		goto label_io_error;
+	}
+	if (err)
+		g_clear_error(&err);
```
fvennetier: `err` won't be set by `g_file_set_contents` if it returns true.

PR #1396, opened 2018-03-21
Parallelize the rebuilding of the chunks

AymericDu:


oio/blob/rebuilder.py
```diff
@@ -20,48 +20,29 @@
...
-            conf.get('allow_same_rawx'))
-        self.input_file = input_file
         self.rdir_client = RdirClient(conf, logger=self.logger)
-        self.content_factory = ContentFactory(conf)
+        self.input_file = input_file
         self.try_chunk_delete = try_chunk_delete
         self.beanstalkd_addr = beanstalkd_addr
         self.beanstalkd_tube = conf.get('beanstalkd_tube', 'rebuild')
         self.beanstalk = None
+        self.num_coroutines = int_value(conf.get('num_coroutines'), 1)
```
jkasarherou: A better name could be `workers`, `parallelism`

PR #1405, opened 2018-03-25
openio: Implement 'directory warmup'

jfsmig:
Ask the Meta1 services to prepare the elections about all the prefixes DB.

### DIFF
--- a/oio/cli/admin/directory.py
+++ b/oio/cli/admin/directory.py
@@ -16,6 +16,10 @@
 from logging import getLogger
 from cliff.command import Command
 from oio.common.exceptions import ServiceUnavailable, ClientException
+from oio.common.utils import load_namespace_conf
+from oio.common import green
+from oio.directory.meta0 import generate_prefixes, count_prefixes
+from eventlet import Queue, Timeout, GreenPile
 
 
 class DirectoryCmd(Command):
@@ -71,6 +75,8 @@ class DirectoryInit(DirectoryCmd):
         return parser
 
     def take_action(self, parsed_args):
+        from time import sleep
+
         self.log.debug('take_action(%s)', parsed_args)
         mapping = self.get_prefix_mapping(parsed_args)
         mapping.load()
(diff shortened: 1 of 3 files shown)


oio/directory/meta0.py
```diff
@@ -437,3 +437,32 @@ def rebalance(self, max_loops=65536):
                 svc_bases = self.get_managed_bases(svc)
                 self.logger.info("meta1 %s has %d bases",
                                  svc['addr'], len(svc_bases))
+
+
+def count_prefixes(digits):
+    """Returns the number of real prefixes in meta0/meta1.
+    Raises an exception if the prefix number is not acceptable."""
+    count = {0: 1, 1: 16, 2: 256, 3: 4096, 4: 65536}
```
fvennetier: Could be `16**digits`.

oio/directory/meta0.py
```diff
@@ -437,3 +437,32 @@ def rebalance(self, max_loops=65536):
...
+    if digits in count:
+        return count[digits]
+    raise Exception("Invalid prefix")
```
fvennetier: `ValueError('Invalid number of digits')` would be more pythonic.

oio/directory/meta0.py
```diff
@@ -437,3 +437,32 @@ def rebalance(self, max_loops=65536):
...
+    if digits == 0:
+        return ('',)
+    elif digits == 1:
+        return (c for c in hexa)
+    elif digits == 2:
+        return (''.join((a, b)) for a, b in product(hexa, repeat=2))
+    elif digits == 3:
+        return (''.join((a, b, c)) for a, b, c in product(hexa, repeat=3))
+    elif digits == 4:
+        return (''.join((a, b, c, d)) for a, b, c, d in product(hexa, repeat=4))
```
fvennetier: Since `product` yields tuples, you can simplify like this:
```
if digits == 0:
    return ('',)
else:
    return (''.join(pfx) for pfx in product(hexa, repeat=digits))
```

oio/cli/admin/directory.py
```diff
@@ -175,3 +178,94 @@ def take_action(self, parsed_args):
...
+                prefix_queue.put(prefix)
+                # Display the progression
+                ratio = float(sent) / total
+                if ratio >= trace_next:
+                    self.log.info("... %d%%", int(ratio * 100.0))
+                    trace_next += trace_increment
+
+            self.log.debug("Send the termination marker")
+            for i in range(workers_count):
+                prefix_queue.put(None)
```
AymericDu: You can use [`prefix_queue.task_done()`](https://docs.python.org/2/library/queue.html#Queue.Queue.task_done). And you wait for the end with [`prefix_queue.join()`](https://docs.python.org/2/library/queue.html#Queue.Queue.join).

PR #1411, opened 2018-03-27
oio-account-server: Minimal access log for read-only requests

jfsmig:
Ugly as hell, but we need that quickly.

### DIFF
--- a/oio/account/server.py
+++ b/oio/account/server.py
@@ -17,6 +17,7 @@
 from werkzeug.wrappers import Response
 from werkzeug.routing import Map, Rule
 from werkzeug.exceptions import NotFound, BadRequest, Conflict
+from functools import wraps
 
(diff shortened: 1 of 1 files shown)


oio/account/server.py
```diff
@@ -23,6 +23,21 @@
 from oio.common.wsgi import WerkzeugApp
 
 
+def access_log(func):
+    from functools import wraps
```
fvennetier: No need to delay this import statement.

oio/account/server.py
```diff
@@ -23,6 +23,21 @@
...
+
+    @wraps(func)
+    def _access_log_wrapper(self, *args, **kwargs):
+        from time import time
```
fvennetier: No need to delay this import statement. Also, you may want to import `oio.common.utils.monotonic_time` from 4.x branch.

PR #1413, opened 2018-03-28
oio/event: check value of safe_decode before using it

murlock:


oio/event/consumer.py
```diff
@@ -223,16 +223,22 @@ def handle(self, beanstalk):
...
-                except (ClientException, OioNetworkException) as exc:
+                if not event:
                     self.logger.warn("Burying event %s (%s): %s",
-                                     job_id, event.get('event'), exc)
-                    beanstalk.bury(job_id)
-                except Exception:
-                    self.logger.exception("Burying event %s: %s",
-                                          job_id, event)
+                                     job_id, event.get('event'),
+                                     "malformed")
```
fvennetier: Seems like a _Bad Idea ©_ to call `event.get()` after `if not event`.

jfsmig: Thank you!

PR #1419, opened 2018-03-30
Several modifications in wsgi.py

newtoncorp:


jfsmig:
Rebased and fixed in #1423

PR #1421, opened 2018-03-30
sqliterepo: prevent deadlock during database deletion

fvennetier:


sqliterepo/cache.c
```diff
@@ -673,6 +680,12 @@ sqlx_cache_open_and_lock_base(sqlx_cache_t *cache, const hashstr_t *hname,
 				g_cond_wait_until(wait_cond, &cache->lock,
 						g_get_monotonic_time() + _cache_period_cond_wait);
 				goto retry;
+
+			case SQLX_BASE_CLOSING_FOR_DELETION:
+				err = NEWERROR(CODE_CONTAINER_NOTFOUND,
+						"Base [%s] about to be deleted",
+						hashstr_str(hname));
```
fvennetier: Here we may want to return another code, like `CODE_CONTAINER_FROZEN` (still there, but cannot be modified) or `CODE_UNAVAILABLE` (to make the calling service retry later, e.g. to recreate the container).

PR #1461, opened 2018-05-04
[WIP] Must send content ID to create object

AymericDu:
##### SUMMARY

- `meta2`: Require to send a content ID to create an object
- `rawx` and `oio-blob-indexer`: Send the right container ID and chunk ID to `rdir`
- `rawx`: Use chunk ID as the key for the fullpath
- `rawx`: Use the same content ID for the linked chunks

##### ISSUE TYPE

 - Feature Pull Request
 - Bugfix Pull Request

##### COMPONENT NAME

- Python API
- `meta2`
- `rawx`
- `oio-blob-auditor`
- `oio-blob-indexer`
- `oio_blob-mover`

##### SDS VERSION

```
openio 4.1.23.dev35
```

##### ADDITIONAL INFORMATION

Now, it's possible to rebuild the linked chunk, but the links aren't rebuilt.


oio/common/fullpath.py
```diff
@@ -0,0 +1,29 @@
...
+
+def encode_fullpath(account, container, path, version):
+    return '{0}/{1}/{2}/{3}'.format(quote_plus(account),
+                                    quote_plus(container),
+                                    quote_plus(path),
+                                    version)
+
+
+def decode_fullpath(fullpath):
+    fp = fullpath.split('/')
```
murlock: May be`fullpath.split('/', 4)` or `if len(fp) != 4: raise Exception`  ?

oio/blob/client.py
```diff
@@ -100,10 +115,35 @@ def __delete_chunk(chunk_):
         resps = [resp for resp in pile if resp]
         return resps
 
-    def chunk_link(self, target, link, **kwargs):
-        headers = kwargs.get('headers')
-        headers["Destination"] = link[:-64] + "/" + link[-64:]
-        return self.http_pool.request('COPY', target, headers=headers)
+    def _generate_fullchunk_copy(self, chunk, random_hex=60, **kwargs):
+        rnd = ''.join(random.choice('0123456789ABCDEF')
```
murlock: Maybe we should use a constant instead of duplicate this 0...F value

rawx-apache2/src/rawx_repo_core.c
```diff
@@ -332,59 +332,162 @@ request_load_chunk_info_from_headers(request_rec *request,
...
+		goto end_check_chunk_content_fullpath;
+	}
+	guint8 container_id[32];
+	char container_hexid[65];
+	// NS is unused
+	oio_str_hash_name(container_id, NULL, account, container);
+	oio_str_bin2hex(container_id, sizeof(container_id),
+			container_hexid, sizeof(container_hexid));
+	if (chunk->container_id) {
+		if (apr_strnatcmp(chunk->container_id, container_hexid) != 0) {
```
fvennetier: A regular string comparison should be sufficient.

rawx-lib/src/attr_handler.c
```diff
@@ -364,6 +338,23 @@ get_compression_info_in_attr(const char *p, GError ** error, GHashTable *table)
 	return TRUE;
 }
 
+gboolean
+remove_fullpath_in_attr(const char *p, GError **error, const gchar *hex_chunkid)
```
fvennetier: Remove *from*.

PR #1464, opened 2018-05-15
Support services IDs for meta2 services

murlock:
##### SUMMARY

Meta2 can now be identified with `service_id` 

##### ISSUE TYPE
 - Feature Pull Request

##### COMPONENT NAME
- meta2

##### SDS VERSION

```
openio 4.2.0.0b3.dev2
```


proxy/m2_actions.c
```diff
@@ -26,6 +26,25 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
 #include "common.h"
 #include "actions.h"
 
+
+static char*
```
jfsmig: The return type should be `gchar*` since the return string is allocated with a GLib-2.0 function, at least to give the hint that it must also be freed with a GLib-2.0 function.

proxy/m2_actions.c
```diff
@@ -1099,6 +1119,8 @@ action_m2_container_destroy (struct req_args_s *args)
 			err = m2v2_remote_execute_DESTROY_many(urlv+1, args->url,
 					flag_force);
 		}
+
+		// free(url);
```
jfsmig: Should that commented code be removed ?
If not, `g_free` would be more suitable :)

proxy/m2_actions.c
```diff
@@ -1067,6 +1086,7 @@ action_m2_container_destroy (struct req_args_s *args)
 		const guint32 flag_force = (force) ? M2V2_DESTROY_FORCE : 0;
 
 		meta1_urlv_shift_addr(urlv);
+		// char *url = _resolve_service_id(urlv[0]);
```
jfsmig: Should that commented code be removed ?

proxy/m2_actions.c
```diff
@@ -1357,11 +1379,13 @@ _m2_container_snapshot(struct req_args_s *args, struct json_object *jargs)
 
 	meta1_urlv_shift_addr(urlv);
 	CLIENT_CTX(ctx, args, type, 1);
+	char *url = _resolve_service_id(urlv[0]);
```
jfsmig: That function seems to return a GLib-2.0 allocated string. Using a `gchar*` seems better.

proxy/m2_actions.c
```diff
@@ -1357,11 +1379,13 @@ _m2_container_snapshot(struct req_args_s *args, struct json_object *jargs)
...
 	GByteArray * _pack(const struct sqlx_name_s *n) {
-		return sqlx_pack_SNAPSHOT(n, urlv[0], target_cid, seq_num, DL());
+		return sqlx_pack_SNAPSHOT(n, url, target_cid, seq_num, DL());
 	}
 
 	err = _resolve_meta2(args, CLIENT_PREFER_MASTER, _pack, NULL);
+	free(url);
```
jfsmig: Mixing GLib-2.0 and glibc memory allocation functions is generally not a good idea.

proxy/m2_actions.c
fvennetier: I saw it, then forgot. It's fixed now.

PR #1466, opened 2018-05-16
oio-reset: Cleanup the datadir when explicitly configured

jfsmig:
##### SUMMARY
Do not forget to clean the files in the data directory.

Prior to this patch, `~/.oio/sds/data` when cleaned even if the data directory was set to a different value.

##### ISSUE TYPE
`enchancement`

##### COMPONENT NAME
`oio-reset`

##### SDS VERSION
`4.2.0`

### DIFF
--- a/tools/oio-reset.sh
+++ b/tools/oio-reset.sh
@@ -130,7 +130,7 @@ done
 
 mkdir -p "$OIO"
 cd "$OIO"
-rm -rf sds.conf sds/{conf,data,run,logs}
+rm -rf sds.conf sds/{conf,run,logs}
 if [[ -d "$DATADIR" ]] ; then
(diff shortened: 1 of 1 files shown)


tools/oio-reset.sh
```diff
@@ -118,7 +128,13 @@ done
 
 # Generate a new configuration and start the new gridinit
 
-mkdir -p "$OIO" && cd "$OIO" && (rm -rf sds.conf sds/{conf,data,run,logs})
+mkdir -p "$OIO"
+cd "$OIO"
+rm -rf sds.conf sds/{conf,data,run,logs}
```
fvennetier: `~/.oio/sds/data` is still cleaned.

PR #1470, opened 2018-05-23
tools: fix oio-reset.sh with invalid bootstrap

murlock:
##### SUMMARY

Fix bootstrap of sds:

Restart of meta1 before rdir assignation was not checked before `openio volume boostrap` 

##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT NAME

tools/oio-reset.sh

##### SDS VERSION
```
4.2.x
```

### DIFF
--- a/tools/oio-reset.sh
+++ b/tools/oio-reset.sh
@@ -170,8 +170,13 @@ $cmd_openio directory bootstrap --check \
 
 echo -e "\n### Assign rdir services"
 # Force meta1 services to reload meta0 cache
+for addr in $(oio-test-config.py -t meta1); do
+    $cmd_openio cluster lock meta1 $addr
+done
 gridinit_cmd -S "$GRIDINIT_SOCK" restart "@meta1" >/dev/null
-sleep 1
+COUNT=$(oio-test-config.py -c -t meta1)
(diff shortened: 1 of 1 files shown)


tools/oio-reset.sh
```diff
@@ -170,8 +170,13 @@ $cmd_openio directory bootstrap --check \
 
 echo -e "\n### Assign rdir services"
 # Force meta1 services to reload meta0 cache
+for addr in $(oio-test-config.py -t meta1); do
+    $cmd_openio cluster lock meta1 $addr
+done
```
fvennetier: I'm not sure this is useful, but it seems harmless, so it's ok.

tools/oio-reset.sh
```diff
@@ -170,8 +170,13 @@ $cmd_openio directory bootstrap --check \
...
 gridinit_cmd -S "$GRIDINIT_SOCK" restart "@meta1" >/dev/null
-sleep 1
+COUNT=$(oio-test-config.py -c -t meta1)
+$cmd_openio cluster wait -d 30 -u -n "$COUNT" meta1
```
fvennetier: :+1:

PR #1483, opened 2018-06-04
Set properties when linking objects

murlock:
##### SUMMARY

Metadata was copied from original source without possibilities to change them

It is first part that fix modifying ACL of existing object on Swift / S3.

Like S3/Swift, we allow to copy or replace metadata through a dedicated parameter.

##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT NAME

OIO

##### SDS VERSION
```
4.1.x
```

##### ADDITIONAL INFORMATION

Behavior to adopt is not clear as API must allow to specify metadata or not (https://developer.openstack.org/api-ref/object-store/index.html#copy-object with `X-Fresh-Metadata` or https://docs.aws.amazon.com/AmazonS3/latest/API/RESTObjectCOPY.html with `x-amz-metadata-directive`)

And it should not break container snapshot of `4.2`

### DIFF
--- a/oio/api/object_storage.py
+++ b/oio/api/object_storage.py
@@ -935,7 +935,7 @@ class ObjectStorageApi(object):
     @patch_kwargs
     def object_fastcopy(self, target_account, target_container, target_obj,
                         link_account, link_container, link_obj,
-                        version=None,  **kwargs):
+                        version=None, properties_directive='COPY', **kwargs):
         meta, chunks = self.object_locate(
             target_account, target_container, target_obj, version=version,
             **kwargs)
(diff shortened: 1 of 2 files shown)


oio/api/object_storage.py
```diff
@@ -951,6 +951,14 @@ def object_fastcopy(self, target_account, target_container, target_obj,
                                            link_obj, version)
         data = {'chunks': chunks_copies,
                 'properties': meta["properties"] or {}}
+        if properties_directive == 'REPLACE':
+            if 'metadata' in kwargs:
+                # TODO it should emit a DeprecationWarning
```
fvennetier: Yes.

PR #1484, opened 2018-06-06
Improve the throughput of the C client when uploading

jfsmig:
##### SUMMARY
* Reduce the number of memory copies
* Work with larger buffers
* Only reload data when necessary
* Fix the management of timeouts, previously polled too often

##### ISSUE TYPE
`enhancement`

##### COMPONENT NAME
`core`

##### SDS VERSION
`openio 4.1.24`

### DIFF
--- a/core/http_put.c
+++ b/core/http_put.c
@@ -627,8 +627,8 @@ http_put_step (struct http_put_s *p)
 		long timeout = 0;
 		curl_multi_timeout (p->mhandle, &timeout);
 		if (timeout < 1000) {
-			tv.tv_sec = 0;
-			tv.tv_usec = timeout * 1000;
+			tv.tv_sec = timeout / 1000;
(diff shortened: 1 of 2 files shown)


core/http_put.c
```diff
@@ -627,8 +627,8 @@ http_put_step (struct http_put_s *p)
 		long timeout = 0;
 		curl_multi_timeout (p->mhandle, &timeout);
```
murlock: Case with timeout == -1 is unmanaged

core/http_put.c
```diff
@@ -627,8 +627,8 @@ http_put_step (struct http_put_s *p)
...
 		if (timeout < 1000) {
-			tv.tv_sec = 0;
-			tv.tv_usec = timeout * 1000;
+			tv.tv_sec = timeout / 1000;
```
murlock: if timeout < 1000, then timeout / 1000 is still 0 ?

PR #1493, opened 2018-06-12
sqliterepo/election: Fix the detection of Rogue nodes

jfsmig:
##### SUMMARY
Fix the detection of Rogue Nodes in the FSM of the elections in sqliterepo.

Prior to the patch, a rogue node caused the election member to never reach the MASTER state, without removing the rogue node.

##### ISSUE TYPE
`bugfix`

##### COMPONENT NAME
`sqliterepo`

##### SDS VERSION
`openio 4.1.25.dev9`


jfsmig:
Introduced at commit ecc099904d0af75f02515ff34b545a4d92d0d093

PR #1494, opened 2018-06-12
meta*, sqlx: Restore the cache of the resolver

jfsmig:
##### SUMMARY
In terms of performance, bring the 4.2.x in the row of the 4.1.x

##### ISSUE TYPE
`enhancement`

##### COMPONENT NAME
`sqliterepo`, `resolver`

##### SDS VERSION
`openio 4.1.25.dev45`

### DIFF
--- a/sqlx/sqlx_service.c
+++ b/sqlx/sqlx_service.c
@@ -362,7 +362,11 @@ _init_configless_structures(struct sqlx_service_s *ss)
 		GRID_WARN("SERVICE init error: memory allocation failure");
 		return FALSE;
 	}
-	oio_var_fix_one("resolver.cache.enabled", "false");
+
+	/* JFS: It was a good idea initially, to turn caching of and avoid
+	 * many decache orders to have predictive movers and rebuilders...
+	 * But it turns so slow without that cache :/ */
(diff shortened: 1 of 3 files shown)


sqlx/sqlx_service.c
```diff
@@ -358,7 +358,11 @@ _init_configless_structures(struct sqlx_service_s *ss)
 		GRID_WARN("SERVICE init error: memory allocation failure");
 		return FALSE;
 	}
-	oio_var_fix_one("resolver.cache.enabled", "false");
+
+	/* JFS: It was a good idea initially, to turn caching of and avoid
+	 * many decache orders to have predictive movers and rebuilders...
+	 * But it turns so slow without that cache :/ */
+	//oio_var_fix_one("resolver.cache.enabled", "false");
```
fvennetier: Could we make `resolver.cache.enabled` false by default*, and enable it by configuration when someone complains about speed? I'm not sure trading reliability for speed is a good idea.

_\* oio-proxy has another parameter overriding this one, and it will continue using cache_

jfsmig: You are right. Let's do this.

PR #1499, opened 2018-06-22
Do not swallow account exceptions

fvennetier:
##### SUMMARY
Before this commit, for any kind of error, the Python API was responding "No Account service found". Now it will at least log the initial error.

##### ISSUE TYPE
Bugfix Pull Request

##### COMPONENT NAME
Python API

##### SDS VERSION
```
openio 4.1.27.dev6
```


oio/account/client.py
```diff
@@ -75,7 +72,7 @@ def _maybe_refresh_endpoint(self):
                         raise
                     self.logger.warn(
                             "Failed to refresh account endpoint: %s", exc)
-                except ClientException:
+                except OioException:
```
vdombrovski: I think you forgot to import OioException here, aswell as remove ClientException import from the top of the file as it is not used anymore.

fvennetier: No, that's fine :wink:

PR #1501, opened 2018-06-25
tests: wait longer for events to be processed (temporary fix)

fvennetier:
##### SUMMARY
We have seen an increase in event processing time. Until we find the problem, increase the wait time in tests using events.

##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT NAME
Tests


murlock:
Replaced by #1506

PR #1504, opened 2018-06-27
Allow 99 max line length in python code

jkasarherou:
##### SUMMARY

Allow 99 max line length in python code (previously set to 79).


fvennetier:
I'm not very fond of long lines.

jkasarherou:
> "long"

PR #1510, opened 2018-07-02
events: add webhook filter

jkasarherou:
##### SUMMARY
Add a webhook filter to event-agent.

##### ISSUE TYPE

 - Feature Pull Request

##### COMPONENT NAME

oio-event-agent

##### ADDITIONAL INFORMATION
Sends HTTP POST to configured endpoint in filter.
Sample
```
{
  "eventId": "17",
  "eventType": "storage.content.new",
  "timestamp": 1530510293897297,
  "data": {
    "id": "025DE780FB6F050045A4AAFE741AAA9D",
    "account": "myaccount",
    "container": "1",
    "name": "magic",
    "md5Hash": "4561FADE4C4FC48E526E72871C470EEC",
    "contentType": "application/octet-stream",
    "policy": "SINGLE",
    "chunkMethod": "plain/nb_copy=1",
    "size": 24301,
    "creationTime": 1530510293,
    "modificationTime": 1530510293,
    "version": 1530510293866760,
    "metadata": {
      "foo": "bar"
    }
  }
}
```

### DIFF
--- a/oio/event/filters/webhook.py
+++ b/oio/event/filters/webhook.py
@@ -1,0 +1,94 @@
+# Copyright (C) 2015-2018 OpenIO SAS, as part of OpenIO SDS
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Affero General Public License as
+# published by the Free Software Foundation, either version 3 of the
+# License, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+
+from oio.common import exceptions
+from oio.common.http_urllib3 import urllib3, get_pool_manager, \
+    oio_exception_from_httperror
+from oio.common.json import json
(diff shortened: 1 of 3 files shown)


oio/event/filters/webhook.py
```diff
@@ -0,0 +1,94 @@
...
+
+    def _request(self, data):
+        try:
+            resp = self.http.request(
+                "POST",
+                self.endpoint,
+                headers={'Content-Type': 'application/json'},
+                body=data)
+            if resp.status >= 400:
+                raise exceptions.from_response(resp, '')
```
fvennetier: Is there a reason to provide an empty string instead of the default `None` for the `body` argument?

jkasarherou: Will fix this

oio/event/filters/webhook.py
```diff
@@ -0,0 +1,94 @@
...
+                raise exceptions.from_response(resp)
+        except urllib3.exceptions.HTTPError as exc:
+            oio_exception_from_httperror(exc)
+
+    def process(self, env, cb):
+        event = Event(env)
+
+        url = env['url']
+        alias = extract_from_event('aliases', env)
+        content_header = extract_from_event('contents_headers', env)
```
fvennetier: Not all event types contain aliases or content headers. I think the `process` method should check `event.event_type` before trying to do anything else.

jkasarherou: Instead of checking the event_type (and having to maintain the supported event type list), how about silently dropping the incomplete events?

PR #1526, opened 2018-07-18
CLI: Allow CID for object and container commands

acornard:
##### SUMMARY
Allow use CID instead of container name to openio object and container command
Fix OB-143

##### ISSUE TYPE
 - Feature Pull Request

##### COMPONENT NAME
CLI
##### SDS VERSION
```
openio 4.2.1.dev25
```

### DIFF
--- a/oio/cli/container/container.py
+++ b/oio/cli/container/container.py
@@ -299,6 +299,13 @@ class FlushContainer(ContainerCommandMixin, command.Command):
  on the event system"""
         )
         self.patch_parser_container(parser)
+        parser.add_argument(
+            '--quickly',
+            action='store_true',
+            dest='quick',
+            help="""Flush container quickly, may put high pressure
+ on the event system"""
+)
         return parser
 
     def take_action(self, parsed_args):
--- a/oio/cli/container/container.py
+++ b/oio/cli/container/container.py
@@ -67,6 +67,7 @@ class SetPropertyCommandMixin(object):
 """
         )
 
+
 class ContainerCommandMixin(object):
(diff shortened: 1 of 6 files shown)


oio/cli/container/container.py
```diff
@@ -253,38 +299,41 @@ def get_parser(self, prog_name):
...
         self.log.debug('take_action(%s)', parsed_args)
-
+        super(FlushContainer, self).take_action_container(parsed_args)
+        if parsed_args.cid is not None:
+            data = self.app.client_manager.storage.container_get_properties(
+                self.app.client_manager.account,
+                parsed_args.container,
+                cid=parsed_args.cid
+            )
+            parsed_args.container = data['system']['sys.user.name']
```
fvennetier: This pattern (getting container name from a CID) is repeated several times. It deserves a dedicated method, like `resolve_cid`, which would return an account and container name (as a tuple or a dict), by requesting meta1 (here we are requesting meta2). This could be done in another PR, though.

oio/cli/container/container.py
```diff
@@ -120,21 +169,17 @@ def take_action(self, parsed_args):
...
 
 
-class SetContainer(SetPropertyCommandMixin, command.Command):
-    """
-    Set container properties, quota, storage policy, status or versioning.
-    """
+class SetContainer(SetPropertyCommandMixin,
+                   ContainerCommandMixin, command.Command):
+    """Set container properties, quota, storage policy,
+ status or versioning."""
```
fvennetier: Indentation seems weird.

PR #1529, opened 2018-07-24
Add options to blob-mover

acornard:
##### SUMMARY

Add options to `oio-blob-mover`.

##### ISSUE TYPE

 - Feature Pull Request

##### COMPONENT NAME

- `oio-blob-mover`

##### SDS VERSION

```
openio 4.2.2.dev4
```

##### ADDITIONAL INFORMATION

The mover can be used as a command line.

To run mover as daemon, you can specify `--daemon` option with the interval.
You can also move only some chunk with `--limit` option.
You can specify disk usage target with  `--usage-target` option.

`--generate-config` can be used to edit configuration file or create a new one with the values of other options.

### DIFF
--- a/bin/oio-blob-mover
+++ b/bin/oio-blob-mover
@@ -21,29 +21,94 @@ import argparse
 from oio.blob.mover import BlobMover
 from oio.common.daemon import run_daemon
 
+
 def make_arg_parser():
     parser = argparse.ArgumentParser()
-    parser.add_argument('config', help="Configuration file")
-    parser.add_argument('--number', type=int,
-                        help="Number of chunk to move")
+    parser.add_argument('--namespace', '--ns', help="Namespace")
+    parser.add_argument('--volume', help="Volume to move")
+    parser.add_argument('config', help="Configuration file or empty file")
+    parser.add_argument('--generate-config', action='store_true',
+                        help="Generate configuration file with given arguments")
+    parser.add_argument('--edit-config', action='store_true',
+                        help="Edit configuration file with given arguments")
+    parser.add_argument('--limit', type=int,
+                        help="Number of chunks to move")
     parser.add_argument('--usage-target', type=int,
-                        help="Percentage of disk usage target")
-    parser.add_argument('--daemon', type=int,
-                        help="Interval to run the mover in seconds.")
+                        help="Target percentage of volume usage")
+    parser.add_argument('--daemon', action='store_true',
+                        help="Run mover as a daemon.")
+    parser.add_argument('--report-interval', type=int,
+                        help="Interval between passes in seconds")
+    parser.add_argument('--usage-check-interval', type=int,
+                        help="Interval between disk usage check in seconds")
     parser.add_argument('--verbose', '-v', action='store_true',
                         help="More verbose output")
+    parser.add_argument('--user', help='run daemon as user')
+    parser.add_argument('--bytes-per-second', type=int,
+                        help="Throttle: max byte per second")
+    parser.add_argument('--chunks-per-second', type=int,
+                        help="Throttle: max chunks per second")
+    levels = ['DEBUG', 'INFO', 'WARN', 'ERROR']
+    parser.add_argument('--log-level', choices=levels,
+                        help="Log level")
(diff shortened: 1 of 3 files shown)


bin/oio-blob-mover
```diff
@@ -16,12 +16,34 @@
...
 from oio.blob.mover import BlobMover
 from oio.common.daemon import run_daemon
-from oio.common.configuration import parse_options
-from optparse import OptionParser
+
+def make_arg_parser():
+    parser = argparse.ArgumentParser()
+    parser.add_argument('config', help="Configuration file")
+    parser.add_argument('--number', type=int,
+                        help="Number of chunk to move")
```
fvennetier: Number of chunk**s** to move.

bin/oio-blob-mover
```diff
@@ -16,12 +16,34 @@
...
+    parser.add_argument('--usage-target', type=int,
+                        help="Percentage of disk usage target")
```
fvennetier: Target percentage of volume usage

bin/oio-blob-mover
```diff
@@ -16,12 +16,34 @@
...
+    parser.add_argument('--daemon', type=int,
+                        help="Interval to run the mover in seconds.")
```
fvennetier: This is ambiguous. The option is named "daemon" but we ask for a duration.
I think it would be clearer to make two arguments:
- a boolean to make the process a daemon;
- an integer to change the duration between passes (with a good default value).

bin/oio-blob-mover
```diff
@@ -16,12 +16,34 @@
...
+import argparse
+
 from oio.blob.mover import BlobMover
 from oio.common.daemon import run_daemon
-from oio.common.configuration import parse_options
-from optparse import OptionParser
+
+def make_arg_parser():
+    parser = argparse.ArgumentParser()
+    parser.add_argument('config', help="Configuration file")
```
fvennetier: Still using a configuration file for compatibility with previous versions is a good think, but now there are command line parameters, requiring it is a bit weird, I think.

oio/blob/mover.py
```diff
@@ -65,14 +64,17 @@ def __init__(self, conf, logger, volume):
         self.container_client = ContainerClient(conf, logger=self.logger)
         self.content_factory = ContentFactory(conf)
 
-    def mover_pass(self):
+    def mover_pass(self, usage_target=None, number=None, **kwargs):
```
fvennetier: I always though we could not use "number" as a variable name.

bin/oio-blob-mover
```diff
@@ -16,12 +16,99 @@
...
+    daemon = args.daemon
+    verbose = args.verbose
+    config = args.config
+    if args.generate_config:
+        config = generate_config_file(config, 'w')
+    if args.edit_config:
+        config = generate_config_file(config, 'a')
+    run_daemon(BlobMover, config, limit=limit,
+               usage_target=usage_target, daemon=daemon,
+               verbose=verbose)
```
AymericDu: `limit` and` usage_target` should be added to the `conf`. That way, you only manage the configuration in the constructor in `BlobMoverWorker`.

bin/oio-blob-mover
```diff
@@ -16,12 +16,99 @@
...
+    def create_content():
+        header = "\n[blob-mover]\n"
+        cont = dict()
+        add_value(cont, 'namespace', args.namespace)
+        add_value(cont, 'volume', args.volume)
+        add_value(cont, 'log_level', args.log_level)
+        add_value(cont, 'log_facility', args.log_facility)
+        add_value(cont, 'log_address', args.log_address)
+        add_value(cont, 'syslog_prefix', args.log_syslog_prefix)
+        add_value(cont, 'usage_target', args.usage_target)
```
AymericDu: You forgot `limit`.

oio/common/utils.py
```diff
@@ -88,9 +88,19 @@ def paths_gen(volume_path):
 
 def statfs(volume):
```
murlock: Function deserve a comment

PR #1534, opened 2018-08-07
Rebalance rawx

acornard:
##### SUMMARY
Add script used to rebalance rawx

##### ISSUE TYPE
 - Feature Pull Request

##### SDS VERSION
```
openio 4.2.2.dev8
```

### DIFF
--- a/tools/oio-rebalance-rawx.py
+++ b/tools/oio-rebalance-rawx.py
@@ -79,19 +79,23 @@ def main():
     for addr in dic:
         infos_srv = {"addr": addr, "type": "rawx"}
         print("Lock rawx at " + addr)
-        cs.lock_score(infos_srv)
-        print("Run mover on rawx at " + addr)
-        try:
-            output = move_rawx(dic[addr], target_use,
-                               args.user, args.namespace, conf)
-            if(args.dry_run):
+        if not args.dry_run:
+            cs.lock_score(infos_srv)
+        print("Run mover on rawx at " + addr +
+              " to get disk usage under " + str(target_use))
+        if not args.dry_run:
+            try:
+                output = move_rawx(dic[addr], target_use,
+                                   args.user, args.namespace, conf)
                 print(output)
-        except Exception as err:
-            print("ERROR: " + str(err))
+            except Exception as err:
+                print("ERROR: " + str(err))
         print("Unlock rawx at " + addr)
-        cs.unlock_score(infos_srv)
+        if not args.dry_run:
+            cs.unlock_score(infos_srv)
     # Delete mover configuration file
-    remove(conf)
+    if not args.dry_run:
+        remove(conf)
 
 
 if __name__ == '__main__':
--- a/tools/oio-rebalance-rawx.py
+++ b/tools/oio-rebalance-rawx.py
@@ -1,0 +1,89 @@
+#!/usr/bin/env python
+
+# Copyright (C) 2015-2018 OpenIO SAS, as part of OpenIO SDS
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Affero General Public License as
+# published by the Free Software Foundation, either version 3 of the
+# License, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+import argparse
+import shlex
+import subprocess
+from oio.common.json import json as jsonlib
+
+def execute(cmd, stdin=None):
+    """Executes command."""
+    cmdlist = shlex.split(cmd)
+    result = ''
+    result_err = ''
+    print(cmd)
+    stdout = subprocess.PIPE
+    stderr = subprocess.PIPE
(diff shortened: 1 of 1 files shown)


tools/oio-rebalance-rawx.py
```diff
@@ -0,0 +1,89 @@
...
+def json_loads(data):
+    try:
+        return jsonlib.loads(data)
+    except ValueError:
+        raise
+
+def make_arg_parser():
+    parser = argparse.ArgumentParser()
+    parser.add_argument('user', help="System user")
+    parser.add_argument('namespace', help="Namespace")
```
murlock: Please add a `--dry-run` option to show which `rawx` should be rebalanced

tools/oio-rebalance-rawx.py
```diff
@@ -0,0 +1,89 @@
...
+    dic = dict()
+    sum_size = 0
+    nrawx = 0
+    for rawx in data:
+        addr = rawx['Addr']
+        stats = rawx['Stats'].split(' ')
+        volume = rawx['Volume']
+        du = float(stats[2].split('=')[1])
+        # Keep only rawx with big disk usage
+        if nrawx < len(data)/2:
```
murlock: I'm not very fond of always selecting half RAWX to rebalance but for first version, I think it is ok

tools/oio-rebalance-rawx.py
```diff
@@ -0,0 +1,89 @@
...
+# License, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
```
murlock: You should add 
`from __future__ import print_function`

tools/oio-rebalance-rawx.py
```diff
@@ -0,0 +1,89 @@
...
+    result_err = ''
+    print(cmd)
+    stdout = subprocess.PIPE
+    stderr = subprocess.PIPE
+    in_ = subprocess.PIPE if stdin else None
+    proc = subprocess.Popen(cmdlist, stdin=in_, stdout=stdout, stderr=stderr)
+    result, result_err = proc.communicate(stdin)
+    result = result.decode('utf-8')
+    if proc.returncode != 0:
+        print result_err
```
murlock: and use `print(result_err)`

tools/oio-rebalance-rawx.py
```diff
@@ -0,0 +1,89 @@
...
+        # Keep only rawx with big disk usage
+        if nrawx < len(data)/2:
+            dic[du] = (addr, volume)
+        sum_size += du
+        nrawx += 1
+    av = sum_size/nrawx
+    # Lock rawx, run blob-mover and unlock rawx
+    for du in dic:
+            openio('cluster lock rawx ' + dic[du][0])
+            output = execute('oio-blob-mover mover-conf.cfg --generate-config ' +
```
murlock: In case of exception in execute (or on user `Control-C`), current RAWX will be left `locked`: goal is to smoothly rebalance so we have to unlock in every case.

tools/oio-rebalance-rawx.py
```diff
@@ -0,0 +1,89 @@
...
+                    '--user ' + args.user + ' --ns ' + args.namespace +
+                    ' --usage-target ' + str(int(av))+ ' --volume ' +
+                    dic[du][1] + ' -v')
+            print(output)
```
murlock: Since stdout is using PIPE, process will be seen as black box, displaying output only at end of process.

tools/oio-rebalance-rawx.py
```diff
@@ -0,0 +1,89 @@
...
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+import argparse
+import shlex
+import subprocess
+from oio.common.json import json as jsonlib
+
+def execute(cmd, stdin=None):
```
fvennetier: This function already exists in [tests/functional/cli/__init__.py](tests/functional/cli/__init__.py). Maybe we can move it to [oio/cli/__init__.py](oio/cli/__init__.py) to avoid duplicating it.

tools/oio-rebalance-rawx.py
```diff
@@ -0,0 +1,89 @@
...
+def make_arg_parser():
+    parser = argparse.ArgumentParser()
+    parser.add_argument('user', help="System user")
+    parser.add_argument('namespace', help="Namespace")
+    return parser
+
+def main():
+    args = make_arg_parser().parse_args()
+    output = openio('cluster list rawx --stats --format json')
+    data = json_loads(output)
```
fvennetier: For that purpose, we would better call `ConscienceClient.all_services('rawx', full=True)`.

tools/oio-rebalance-rawx.py
```diff
@@ -0,0 +1,98 @@
...
+    for addr in dic:
+        infos_srv = {"addr": addr, "type": "rawx"}
+        print("Lock rawx at " + addr)
+        cs.lock_score(infos_srv)
+        print("Run mover on rawx at " + addr)
+        try:
+            output = move_rawx(dic[addr], target_use,
+                               args.user, args.namespace, conf)
+            if(args.dry_run):
+                print(output)
```
fvennetier: "Dry run" means "do not do any action, just print what you would do in normal mode". Here the action is executed anyway.

PR #1538, opened 2018-08-16
Update blob-auditor usage

acornard:
##### SUMMARY
Change blob-auditor usage

##### ISSUE TYPE
 - Feature Pull Request

##### COMPONENT NAME
oio-blob-auditor

##### SDS VERSION
```
openio 4.2.2.dev37
```

##### ADDITIONAL INFORMATION

The auditor can be used as a command line.

To run auditor as daemon, you can specify `--daemon` option .
`--generate-config` can be used to edit configuration file or create a new one with the values of other options.

### DIFF
--- a/bin/oio-blob-auditor
+++ b/bin/oio-blob-auditor
@@ -16,12 +16,87 @@
 # You should have received a copy of the GNU Affero General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
+import argparse
+
 from oio.blob.auditor import BlobAuditor
 from oio.common.daemon import run_daemon
-from oio.common.configuration import parse_options
-from optparse import OptionParser
+
+
+def make_arg_parser():
+    parser = argparse.ArgumentParser()
+    parser.add_argument('--namespace', '--ns', help="Namespace")
+    parser.add_argument('--volume', help="Volume to audit")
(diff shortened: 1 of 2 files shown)


bin/oio-blob-auditor
```diff
@@ -16,12 +16,87 @@
...
+def main():
+    args = make_arg_parser().parse_args()
+    daemon = args.daemon
+    verbose = args.verbose
+    config = args.config
+
+    if args.generate_config:
+        config = generate_config_file(config, 'w')
+    if args.edit_config:
+        config = generate_config_file(config, 'a')
```
murlock: It seems dangerous since we don't know how duplicate entries are managed ?
(and from PoV of devops, it seems always a bad idea to have duplicate entries)

oio/blob/auditor.py
```diff
@@ -232,13 +232,17 @@ def __init__(self, conf, **kwargs):
         self.volume = volume
 
     def run(self, *args, **kwargs):
-        while True:
+        work = True
+        while work:
             try:
                 worker = BlobAuditorWorker(self.conf, self.logger, self.volume)
                 worker.audit_pass()
+                work = False
```
murlock: If an exception occurs without `daemon` mode enabled, `work` still remains `true`, it is by design ? (i.e. we want a full pass)

PR #1546, opened 2018-08-28
rawx: Configure coverage

AymericDu:
##### SUMMARY

Configure coverage

##### ISSUE TYPE

 - Coverage Pull Request

##### COMPONENT NAME

- `rawx`

##### SDS VERSION

```
openio 4.2.2.dev132
```


tools/oio-bootstrap.py
```diff
@@ -1608,6 +1608,15 @@ def main():
             [template_wsgi_service_coverage_start,
              template_wsgi_service_descr,
              template_wsgi_service_coverage_stop])
+        global template_gridinit_rawx
+        template_gridinit_rawx = """
```
fvennetier: This is ugly. Could you define the template in another variable, and make the first variable point to the second when needed?

PR #1553, opened 2018-09-12
rawx: Implement beanstalkd client

AymericDu:
##### SUMMARY

- Implement beanstalkd client
- Remove the dependency to kr/beanstalk

##### ISSUE TYPE

 - Feature Pull Request

##### COMPONENT NAME

- `rawx`

##### SDS VERSION

```
openio 4.2.2.dev135
```

### DIFF
--- a/rawx/notifier.go
+++ b/rawx/notifier.go
@@ -22,8 +22,6 @@ import (
 	"strings"
 	"sync"
 	"time"
-
-	"github.com/kr/beanstalk"
 )
 
 type eventData struct {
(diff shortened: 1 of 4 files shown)


rawx/notifier.go
```diff
@@ -81,38 +75,61 @@ func makeBeanstalkNotifier(endpoint string,
...
-				LogWarning("Notify error: %s", err)
-			}
+			notifier.syncNotify(eventJSON)
 		}
 	}()
 }
 
-func (notifier *beanstalkNotifier) stop() {
+func (notifier *beanstalkNotifier) Stop() {
 	close(notifier.queue)
```
jkasarherou: Careful here no guarantees when you close the channel that an `asyncNotify` is not trying to write in the channel, which will crash the goroutine.

PR #1555, opened 2018-09-17
Reduce the log volume of well-known spammer services

jfsmig:
##### SUMMARY
Reduce the access log volume for sqliterepo-based services and the proxy.

##### ISSUE TYPE
`enhancement`

##### COMPONENT NAME
`sqliterepo`, `oio-proxy`

##### SDS VERSION
`openio 4.2.2.dev70`

### DIFF
--- a/proxy/cache_actions.c
+++ b/proxy/cache_actions.c
@@ -25,11 +25,11 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
 enum http_rc_e
 action_forward_stats (struct req_args_s *args)
 {
+	args->rp->no_access();
+
 	const char *id = OPT("id");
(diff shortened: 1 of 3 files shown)


proxy/cache_actions.c
```diff
@@ -120,12 +124,15 @@ enum http_rc_e
...
 {
 	args->rp->no_access();
+
 	return _reply_success_json (args, oio_var_list_as_json());
 }
 
 static enum http_rc_e
 _set_config (struct req_args_s *args, struct json_object *jargs)
 {
+	args->rp->no_access();
```
fvennetier: Are we sure we don't want to know when configuration changes?

PR #1560, opened 2018-09-20
event: Select the alias correctly

AymericDu:
##### SUMMARY

- Select the alias correctly to load the URL with this alias
- Fix the number of parts

##### ISSUE TYPE

 - Bugfix Pull Request

##### COMPONENT NAME

- `meta2`

##### SDS VERSION

```
openio 4.2.2.dev79
```

### DIFF
--- a/meta2v2/meta2_filters_action_content.c
+++ b/meta2v2/meta2_filters_action_content.c
@@ -98,10 +98,21 @@ _m2b_notify_beans(struct meta2_backend_s *m2b, struct oio_url_s *url,
 	if (!send_chunks) {
 		GSList *non_chunks = NULL;
 		for (GSList *l = beans; l; l = l->next) {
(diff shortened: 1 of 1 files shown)


meta2v2/meta2_filters_action_content.c
```diff
@@ -98,10 +98,15 @@ _m2b_notify_beans(struct meta2_backend_s *m2b, struct oio_url_s *url,
 	if (!send_chunks) {
 		GSList *non_chunks = NULL;
 		for (GSList *l = beans; l; l = l->next) {
-			if (DESCR(l->data) != &descr_struct_CHUNKS)
+			if (DESCR(l->data) == &descr_struct_ALIASES) {
+				if (alias)
+					GRID_WARN("Several aliases in same event!");
```
fvennetier: It would be a good idea to give more information: which alias? which content id? which request id?

PR #1561, opened 2018-09-21
Include object metadata in events for indexing

AymericDu:
##### SUMMARY

Include object metadata in events for indexing:
- `content.new`: Use the properties in events
- `content.update`: Load the properties with the `webhook` filter (`event-agent`)
  * It's a new event type
  * It is sent when changing the properties

##### ISSUE TYPE

 - Feature Pull Request

##### COMPONENT NAME

- `meta2`
- `event-agent`

##### SDS VERSION

```
openio 4.2.2.dev151
```

### DIFF
--- a/oio/event/filters/webhook.py
+++ b/oio/event/filters/webhook.py
@@ -19,12 +19,15 @@ from oio.common.http_urllib3 import urllib3, get_pool_manager, \
     oio_exception_from_httperror
 from oio.common.json import json
 from oio.event.evob import Event, EventError
+from oio.event.consumer import EventTypes
 from oio.event.filters.base import Filter
+from oio.container.client import ContainerClient
 
 
 class WebhookFilter(Filter):
 
     def init(self):
+        self.container_client = ContainerClient(self.conf, logger=self.logger)
         self.endpoint = self.conf.get('endpoint')
         # TODO configure pool manager
         self.http = get_pool_manager()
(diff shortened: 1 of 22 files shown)


oio/event/filters/webhook.py
```diff
@@ -19,11 +19,15 @@
...
 from oio.event.evob import Event, EventError
+from oio.event.consumer import EventTypes
 from oio.event.filters.base import Filter
+from oio.container.client import ContainerClient
 
 
 class WebhookFilter(Filter):
+
     def init(self):
+        self.container_client = ContainerClient(self.conf, logger=self.logger)
```
fvennetier: This lets me think that we must review how we create connection pools in the event-agent. We would better share a big one with a known configuration instead of instantiating several with default configuration.

oio/event/filters/webhook.py
```diff
@@ -67,8 +67,29 @@ def process(self, env, cb):
...
                 'modificationTime': alias['mtime'],
                 'version': alias['version'],
-                'metadata': {},
-            },
+                'metadata': properties,
+            })
+        else:
+            all_properties = self.container_client.content_get_properties(
+                account=url['account'], reference=url['user'],
+                content=url['content'])
```
fvennetier: Isn't there a version?

tools/oio-bootstrap.py
```diff
@@ -824,6 +828,9 @@ def save_coverage():
 [filter:volume_index]
 use = egg:oio#volume_index
 
+[filter:webhook]
+use = egg:oio#webhook
```
fvennetier: We should also update `etc/event-handlers.conf-sample`.

oio/event/filters/webhook.py
AymericDu: In the URL, there is no version.

Here is an example:
```
"url":{"ns":"OPENIO","account":"myaccount","user":"toto","path":"magic","content":"EEC419C19E76050065243D428D5FFA00","id":"8B78B3245B74710F3ACC1BEF4978E621F5E764E01FFB5621D23C4EECA2B7BB3D"}
```

Also, I already have the content ID. So the version is useless here.

PR #1579, opened 2018-10-22
lifecycle: Check the S3 tags instead of checking the properties 

AymericDu:
##### SUMMARY

Check the S3 tags instead of checking the properties when applying Lifecycle

##### ISSUE TYPE

 - Feature Pull Request

##### COMPONENT NAME

- `lifecycle`
- `CLI`

##### SDS VERSION

```
openio 4.2.4.dev12
```

### DIFF
--- a/oio/cli/object/object.py
+++ b/oio/cli/object/object.py
@@ -19,6 +19,7 @@ from cliff import command, lister, show
 from eventlet import GreenPool
 from oio.common.http_urllib3 import get_pool_manager
 from oio.common.utils import depaginate
+from oio.common.json import json as jsonlib
 from oio.common import exceptions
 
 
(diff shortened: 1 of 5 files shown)


oio/cli/object/object.py
```diff
@@ -389,6 +395,14 @@ def take_action(self, parsed_args):
         if parsed_args.auto:
             container = self.flatns_manager(obj)
         properties = parsed_args.property
+        if parsed_args.tag:
+            from oio.container.lifecycle import TAGGING_KEY
+            tags = '<Tagging><TagSet>'
+            for k, v in parsed_args.tag.iteritems():
+                tags += '<Tag><Key>%s</Key><Value>%s</Value></Tag>' % (k, v)
+            tags += '</TagSet></Tagging>'
```
fvennetier: This probably builds a good enough XML document, but unfortunately it will override any previous tagging information. I see two solutions:
- Clearly define this behavior in the help of the command line.
- Use another method to build the XML. For example we can load the existing document, parse it with etree (not in the CLI module), and patch it with the new or deleted tags.

PR #1581, opened 2018-10-23
meta2: fix memory leak

murlock:
##### SUMMARY

Fix memory leak

##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT NAME
meta2

##### SDS VERSION
```
master
```

### DIFF
--- a/meta2v2/meta2_utils.c
+++ b/meta2v2/meta2_utils.c
@@ -275,7 +275,7 @@ _load_chunk_quality(GHashTable *qualities, struct bean_PROPERTIES_s *prop)
 			g_hash_table_insert(qualities,
 					g_strdup(key->str + strlen(OIO_CHUNK_SYSMETA_PREFIX)), item);
 		} else {
-			g_free(item);
+			oio_lb_selected_item_free(item);
 		}
 	}
 
@@ -2400,7 +2400,7 @@ m2db_check_content_quality(struct m2v2_sorted_content_s *sorted_content,
 	/* keys are chunk URLs,
 	 * values are <struct oio_lb_selected_item_s *> */
 	GHashTable *chunk_items = g_hash_table_new_full(g_str_hash, g_str_equal,
-			g_free, g_free);
+			g_free, (GDestroyNotify) oio_lb_selected_item_free);
 	_load_chunk_qualities(chunk_items, chunk_meta);
 
 	gboolean _on_metachunk(gpointer ppos, GSList *chunks, gpointer udata UNUSED) {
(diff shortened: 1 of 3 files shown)


meta2v2/meta2_utils.c
```diff
@@ -275,6 +275,8 @@ _load_chunk_quality(GHashTable *qualities, struct bean_PROPERTIES_s *prop)
 			g_hash_table_insert(qualities,
 					g_strdup(key->str + strlen(OIO_CHUNK_SYSMETA_PREFIX)), item);
 		} else {
+			g_free((char *)item->expected_slot);
+			g_free((char *)item->final_slot);
```
AymericDu: If `meta2_json_fill_item_quality` returns an error, `expected_slot` and `final_slot` are not loaded: https://github.com/open-io/oio-sds/blob/master/meta2v2/meta2_utils_json_in.c#L297

meta2v2/meta2_utils.c
```diff
@@ -2445,8 +2447,15 @@ m2db_check_content_quality(struct m2v2_sorted_content_s *sorted_content,
...
+
+	void _on_items(gpointer key UNUSED, struct oio_lb_selected_item_s *item, gpointer data UNUSED) {
+		g_free((char *) item->expected_slot);
+		g_free((char *) item->final_slot);
+	}
+
 	g_tree_foreach(sorted_content->metachunks,
 			(GTraverseFunc)_on_metachunk, NULL);
+	g_hash_table_foreach(chunk_items, (GHFunc)_on_items, NULL);
 	g_hash_table_destroy(chunk_items);
```
AymericDu: You can use the fourth parameter of `g_hash_table_new_full` to free the values automatically: `https://github.com/open-io/oio-sds/blob/master/meta2v2/meta2_utils.c#L2403`

meta2v2/meta2_utils.c
```diff
@@ -2445,8 +2447,15 @@ m2db_check_content_quality(struct m2v2_sorted_content_s *sorted_content,
...
 	}
+
+	void _on_items(gpointer key UNUSED, struct oio_lb_selected_item_s *item, gpointer data UNUSED) {
+		g_free((char *) item->expected_slot);
+		g_free((char *) item->final_slot);
+	}
+
 	g_tree_foreach(sorted_content->metachunks,
 			(GTraverseFunc)_on_metachunk, NULL);
+	g_hash_table_foreach(chunk_items, (GHFunc)_on_items, NULL);
```
fvennetier: We should define `oio_lb_selected_item_free` and pass that to the hash table constructor.

meta2v2/meta2_utils.c
fvennetier: That is not a problem, `g_free` is NULL-safe.

PR #1597, opened 2018-11-12
Fix the purge of objects with a delete marker

AymericDu:
##### SUMMARY

Fix the purge of objects with a delete marker (the last version)

##### ISSUE TYPE

 - Bugfix Pull Request

##### COMPONENT NAME

- `meta2`

##### SDS VERSION

```
openio 4.2.4.dev14
```

### DIFF
--- a/tests/functional/content/test_content_versioning.py
+++ b/tests/functional/content/test_content_versioning.py
@@ -13,8 +13,12 @@
 # You should have received a copy of the GNU Lesser General Public
 # License along with this library.
 
+
+import time
+
 from oio.api.object_storage import ObjectStorageApi
 from tests.utils import BaseTestCase, random_str
+from oio.common.exceptions import NoSuchObject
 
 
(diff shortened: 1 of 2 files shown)


tests/functional/content/test_content_versioning.py
```diff
@@ -213,3 +217,35 @@ def check_num_objects(expected):
...
+        self.api.container.content_purge(
+            self.account, self.container, "versioned")
+        self.assertRaises(NoSuchObject, self.api.object_locate, self.account,
+                          self.container, "versioned")
+        check_num_objects(4)
+
+        system = {'sys.m2.keep_deleted_delay': '1'}
+        self.api.container_set_properties(self.account, self.container,
+                                          system=system)
+        time.sleep(2)
```
fvennetier: Isn't 1s enough?

AymericDu: I tried with 1 second, but the test fails.

PR #1603, opened 2018-11-16
Add a utility to index already existing meta2 databases on disk

aneutron:
##### SUMMARY
- Add a new wrapper to list all meta2 records indexed for a volume.
- Add new meta2 indexing tool, sample conf and tests.

##### ISSUE TYPE
 - Feature Pull Request

##### COMPONENT NAME
`Python API`, `oio-meta2-indexer`

##### SDS VERSION
```
openio 4.3.1.dev47
```

##### ADDITIONAL INFORMATION
- A single process for multiple volumes
- A `Meta2IndexingWorker` is spawned for each volume in a green thread pool.
- Continuous progress logs are emitted in a similar fashion to the `oio-blob-indexer`


oio/directory/indexer.py
```diff
@@ -0,0 +1,257 @@
...
+        if len(self.volumes) == 0:
+            raise exc.OioException("Not meta2 volumes provided to index !")
+        self.volumes = [x.strip() for x in self.volumes]
+        self.pool = ContextPool(len(self.volumes))
+        self.volume_workers = [Meta2IndexingWorker(x, conf) for x in
+                               self.volumes]
+
+    def run(self, *args, **kwargs):
+        for worker in self.volume_workers:
+            self.pool.spawn(run, worker)
```
fvennetier: I think it's possible to pass `Meta2IndexingWorker.run` instead of the "workaround" method.

oio/directory/indexer.py
```diff
@@ -0,0 +1,257 @@
...
+        Possible values of conf relating to this worker are:
+        - interval: (int) in sec time between two full scans. Default: half an
+                    hour.
+        - report_interval: (int) in sec, time between two reports: Default: 300
+        - scanned_per_second: (int) maximum number of indexed databases /s.
+
+        :param volume_path: The volume path to be indexed
+        :param conf: The configuration to be passed to the needed services
+        :param pm: A connection pool manager. If none is given, a new one with
+                a default size of 10 will be created.
```
fvennetier: For this last parameter, I suggest using the same name as everywhere else: `pool_manager`.

oio/directory/indexer.py
```diff
@@ -0,0 +1,257 @@
...
+from oio.common.http_urllib3 import get_pool_manager
+from oio.common.logger import get_logger
+from oio.common.utils import paths_gen
+from oio.directory.client import DirectoryClient
+from oio.rdir.client import RdirClient
+
+import time
+
+
+class Meta2IndexingWorker:
```
fvennetier: I suggest using "new-style" classes only.

oio/directory/indexer.py
```diff
@@ -0,0 +1,257 @@
...
+from oio.common.daemon import Daemon
+from oio.common.easy_value import int_value
+from oio.common.green import ratelimit, ContextPool
+from oio.common.http_urllib3 import get_pool_manager
+from oio.common.logger import get_logger
+from oio.common.utils import paths_gen
+from oio.directory.client import DirectoryClient
+from oio.rdir.client import RdirClient
+
+import time
```
fvennetier: We have had troubles with `time.time()` and `time.sleep()` recently. It's better to import them from `oio.common.green`.

oio/directory/indexer.py
```diff
@@ -0,0 +1,257 @@
...
+        )
+
+    def index_meta2_database(self, db_id):
+        """
+        Add a meta2 database to the rdir index. Fails if the database isn't
+        handled by the current volume.
+
+        :param db_id: The ContentID representing the reference to the database.
+        """
+        if len(db_id) != STRLEN_REFERENCEID:
```
fvennetier: This check will fail when the sequence number of the database will be greater than 9.

oio/directory/indexer.py
```diff
@@ -0,0 +1,257 @@
...
+            return
+        try:
+            srvcs = self.dir_client.list(cid=db_id)
+            account, container = srvcs['account'], srvcs['name']
+            is_peer = self.volume_id in [x['host'] for x in srvcs['srv'] if
+                                         x['type'] == 'meta2']
+            if not is_peer:
+                self.warn("Trying to index a container that isn't handled by"
+                          "this volume", db_id)
+                return
```
fvennetier: In such case, shouldn't we call `meta2_index_delete()`?

oio/directory/indexer.py
```diff
@@ -0,0 +1,257 @@
...
+            cont_url = "{0}/{1}/{2}".format(self.namespace, account, container)
+            self.index_client.meta2_index_push(volume_id=self.volume_id,
+                                               container_url=cont_url,
+                                               mtime=time.time(),
+                                               container_id=container_id)
+
+            self.success_nb += 1
+        except exc.OioException:
+            self.failed_nb += 1
+            self.warn("Unable to to index container", db_id)
```
fvennetier: I suggest also printing the exception message.

oio/directory/indexer.py
```diff
@@ -0,0 +1,257 @@
...
+                self.last_scan_time = time.time()
+                time.sleep(self.scans_interval)
+            except exc.OioException as e:
+                self.logger.exception("ERROR during indexing meta2: %s", e)
+
+    def stop(self):
+        """
+        Could be needed for eventually gracefully stopping.
+        """
+        self.stop = True
```
fvennetier: There is a name conflict between the field and the method.

oio/directory/indexer.py
```diff
@@ -0,0 +1,257 @@
...
+                'index_rate': self.indexed_since_last_report / since_last_rprt,
+                'elapsed': elapsed
+            }
+        )
+        self.last_report_time = now
+        self.indexed_since_last_report = 0
+
+    def warn(self, msg, container_id):
+        self.logger.warn(
+            'volume_id=%(volume_id)s container_id=%(container_id)s %(error)s' %
```
fvennetier: I suggest passing the dictionary to the logging function instead of formatting the string here.

oio/directory/indexer.py
```diff
@@ -0,0 +1,257 @@
...
+        total = self.success_nb + self.failed_nb
+        now = time.time()
+        elapsed = (now - self.start_time) or 0.00001
+        since_last_rprt = (now - self.last_report_time) or 0.00001
+        self.logger.info(
+            'volume_id=%(volume_id)s %(tag)s=%(current_time)s '
+            'elapsed=%(elapsed).02f '
+            'pass=%(pass)d '
+            'errors=%(errors)d '
+            'containers_indexed=%(total_indexed)d %(index_rate).2f/s' % {
```
fvennetier: I suggest passing the dictionary to the logging function instead of formatting the string here.

oio/directory/indexer.py
```diff
@@ -0,0 +1,257 @@
...
+                break
+
+            db_id = db_path.rsplit("/")[-1]
+            db_id = db_id.rsplit(".")
+
+            if len(db_id) != 3:
+                self.warn("Malformed db file name !", db_path)
+                continue
+
+            db_id = ".".join([db_id[0], db_id[1]])
```
vdombrovski: db_id[:2]

oio/directory/indexer.py
```diff
@@ -0,0 +1,257 @@
...
+        self.report("starting")
+
+        for db_path in paths:
+
+            # Graceful exit, hopefully
+            if self.stop:
+                break
+
+            db_id = db_path.rsplit("/")[-1]
+            db_id = db_id.rsplit(".")
```
vdombrovski: db_path.rsplit("/")[-1].rsplit(".")

oio/directory/indexer.py
```diff
@@ -0,0 +1,257 @@
...
+            Default: 300
+        - scanned_per_second: (int) maximum number of indexed databases /s.
+            Default: 1000
+    """
+
+    def __init__(self, conf):
+        super(Meta2Indexer, self).__init__(conf=conf)
+        self.logger = get_logger(conf)
+        if not conf.get("volume_list"):
+            raise exc.OioException("Not meta2 volumes provided to index !")
```
vdombrovski: ~Not~ No

oio/directory/indexer.py
```diff
@@ -0,0 +1,257 @@
...
+        self.volumes = conf.get('volume_list').split(',')
```
vdombrovski: [x.strip for x in conf.get('volume_list', []).split(',')] and remove line 245

oio/directory/indexer.py
```diff
@@ -0,0 +1,257 @@
...
+    """
+    Simple workaround to run the Meta2IndexingWorker in a ContainerPool.
+    :param worker Meta2IndexingWorker The worker to run.
+    """
+    worker.run()
+
+
+class Meta2Indexer(Daemon):
+    """
+    A daemon that spaws a greenlet running a Meta2IndexingWorker
```
vdombrovski: spawns

oio/directory/indexer.py
```diff
@@ -0,0 +1,257 @@
...
+
+        :param db_id: The ContentID representing the reference to the database.
+        """
+        if len(db_id) != STRLEN_REFERENCEID:
+            self.warn('Not a valid container ID', db_id)
+            return
+        try:
+            srvcs = self.dir_client.list(cid=db_id)
+            account, container = srvcs['account'], srvcs['name']
+            is_peer = self.volume_id in [x['host'] for x in srvcs['srv'] if
```
vdombrovski: I don't think you need to check type here, as  srvcs['srv']  contains only M2 anyway.

fvennetier:
I forgot to mention that `oio-meta2-indexer` is missing from `setup.cfg`.

Also, as `oio-blob-mover` and `oio-blob-auditor`, it may be nice to support command-line parameters, at least one to generate a basic configuration file (`-h` does not help).

oio/directory/indexer.py
fvennetier: I disagree with that. We can link arbitrary services to "references".

oio/directory/indexer.py
aneutron: Added it as an option, default to False.

fvennetier:
Also, `oio-meta2-indexer` is still missing from `setup.cfg`.

PR #1609, opened 2018-11-21
Remove index entry when moving a meta2 database with oio-meta2-mover

aneutron:
##### SUMMARY
When moving a meta2 database using the oio-meta2-mover, we have to remove the database from the previous meta2 peer's index.

##### ISSUE TYPE

 - Bugfix Pull Request

##### COMPONENT NAME
`oio-meta2-mover`

##### SDS VERSION

```
openio 4.3.1.dev54
```

##### ADDITIONAL INFORMATION
```
Nil.
```

### DIFF
--- a/oio/directory/meta.py
+++ b/oio/directory/meta.py
@@ -14,6 +14,7 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 from oio.directory.admin import AdminClient
+from oio.rdir.client import RdirClient
 from oio.conscience.client import ConscienceClient
 from oio.common.exceptions import OioException, ServiceBusy
 from oio.common.logger import get_logger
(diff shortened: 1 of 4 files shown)


oio/directory/meta.py
```diff
@@ -118,6 +126,11 @@ def _apply_copy_bases(self, moved, **kwargs):
                         all_peers_ok = False
                 if all_peers_ok:
                     moved_ok.append(base)
+                    if service_type == 'meta2' and kwargs.get('src_service'):
+                        self.rdir.meta2_index_delete(
+                            volume_id=kwargs.get('src_service'),
+                            container_id=cid
+                        )
```
fvennetier: At this point, the old peer it still referenced in the directory. If a crawler walks on it, it will be referenced again. This workaround should be done after `_apply_link_services`.

Also, I would like it to be clear: this is a workaround and definitely not the best solution to deindex the database.

oio/directory/meta1.py
```diff
@@ -53,6 +53,11 @@ def _apply_link_services(self, moved_ok, **kwargs):
                     service_type=service_type, cid=cid,  replace=True,
                     services=dict(host=','.join(peers), type=service_type,
                                   args=args, seq=seq))
+                if service_type == 'meta2' and kwargs.get('src_service'):
+                    self.rdir.meta2_index_delete(
+                        volume_id=kwargs.get('src_service'),
+                        container_id=cid
+                    )
```
fvennetier: I suggest adding a note telling that we will be able to remove this part when either:
- meta1 `account.services` events contain the list of services that have been removed;
- or meta2 sends `storage.container.deleted` events even when called at the sqliterepo level.

aneutron: Added in  f5c557b

PR #1612, opened 2018-11-26
Make an election member emit a DB_USE when becoming MASTER

jfsmig:
##### SUMMARY
Make an election member emit a DB_USE when becoming MASTER. It helps the whole election detect and recover double MASTER conditions: the first/old MASTER that was elected on the basis of a wrong list of members may decide to quit.

##### ISSUE TYPE
`enhancement`, `workaround`

##### COMPONENT NAME
`sqliterepo`

##### SDS VERSION
`openio 4.2.5.dev8`

##### ADDITIONAL INFORMATION
It works by adding a new optional field in an existing message. It breaks nothing.


sqliterepo/repository.c
```diff
@@ -1347,6 +1348,11 @@ sqlx_repository_use_base(sqlx_repository_t *repo, const struct sqlx_name_s *n,
 			err = _base_lazy_recover(repo, n, status);
 		}
 
+		if (status == STEP_MASTER && notify_master) {
+			GRID_WARN("Possible double master");
```
fvennetier: I suggest adding the name of the base to the message.
```suggestion
			GRID_WARN("Possible double master for [%s][%s]", n->base, n->type);
```

PR #1617, opened 2018-11-28
Fix webhook and add tests

murlock:
##### SUMMARY

Webhook was too aggressive on few events, with this PR, webhook will skip retrieving properties on `storage.content.deleted` event.

It also skip `storage.content.append` event due to invalid `content` received and no version available in event data, it is not possible to retrieve properties nor size. 

Fixes OS-235

##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT NAME
- event-handler

##### SDS VERSION
```
4.3.x
```

### DIFF
--- a/tests/functional/event/test_webhook.py
+++ b/tests/functional/event/test_webhook.py
@@ -1,0 +1,107 @@
+# -*- coding: utf-8 -*-
+
+# Copyright (C) 2018 OpenIO SAS, as part of OpenIO SDS
+#
+# This library is free software; you can redistribute it and/or
+# modify it under the terms of the GNU Lesser General Public
+# License as published by the Free Software Foundation; either
+# version 3.0 of the License, or (at your option) any later version.
+#
+# This library is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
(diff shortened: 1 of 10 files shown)


tests/functional/event/test_webhook.py
```diff
@@ -0,0 +1,114 @@
...
+# This library is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+# Lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public
+# License along with this library.
+
+
+import requests
```
fvennetier: Didn't we want to get rid of `requests`, and replace it with `urllib3`?

tests/functional/event/test_webhook.py
```diff
@@ -0,0 +1,114 @@
...
+
+    def _remove(self):
+        self.storage_api.object_delete(self.acct, self.cnt_name, self.obj_name)
+        ret, data = self._get(success=False)
+        self.assertEqual(ret.status_code, requests.codes.not_found)
+        return ret, data
+
+    def test_content_add(self):
+        if not self.conf.get('webhook', ''):
+            self.skipTest('webhook is required')
```
fvennetier: Minor: since all these tests require webhook, these 2 lines can be put in `setUp()`.

PR #1626, opened 2018-12-05
Add User-Agent to the events

AymericDu:
##### SUMMARY

Add User-Agent in the events

##### ISSUE TYPE

 - Feature Pull Request

##### COMPONENT NAME

- `proxy`
- `meta2`

##### SDS VERSION

```
openio 4.2.6.dev11
```

### DIFF
--- a/events/oio_events_queue.c
+++ b/events/oio_events_queue.c
@@ -189,7 +189,12 @@ oio_event__create_with_id(const char *type, struct oio_url_s *url,
 	oio_event__init (gs, type, url);
 	if (request_id && *request_id) {
 		g_string_append_c(gs, ',');
-		oio_str_gstring_append_json_pair(gs, "request_id", request_id);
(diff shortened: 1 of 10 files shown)


events/oio_events_queue.c
```diff
@@ -191,5 +191,10 @@ oio_event__create_with_id(const char *type, struct oio_url_s *url,
 		g_string_append_c(gs, ',');
 		oio_str_gstring_append_json_pair(gs, "request_id", request_id);
 	}
+	const gchar *user_agent = oio_ext_get_user_agent();
+	if (user_agent != NULL) {
+		g_string_append_c(gs, ',');
+		oio_str_gstring_append_json_pair(gs, "user_agent", user_agent);
```
murlock: `user_agent` should be a define, as well as l.192

fvennetier: In the event data, I think that it should be "origin" instead of "user_agent", because it does not represent what has emitted the event, but what is at the origin of the request that triggered the event.

PR #1630, opened 2018-12-12
Check if the meta1 are present when bootstrapping

AymericDu:
##### SUMMARY

Check if the meta1 are present when bootstrapping

##### ISSUE TYPE

 - Bugfix Pull Request

##### COMPONENT NAME

- Python API

##### SDS VERSION

```
openio 4.2.6.dev20
```

### DIFF
--- a/oio/directory/meta0.py
+++ b/oio/directory/meta0.py
@@ -299,6 +299,10 @@ class Meta0PrefixMapping(MetaMapping):
         using `strategy` to find new services.
         """
         self.reset()
+        if len(self.services) < self.replicas:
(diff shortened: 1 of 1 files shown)


oio/directory/meta0.py
```diff
@@ -299,6 +299,8 @@ def bootstrap(self, strategy=None):
         using `strategy` to find new services.
         """
         self.reset()
+        if not self.services:
+            raise ValueError('No meta1 service')
```
fvennetier: We may also check if there are more services than `self.replicas`.

PR #1632, opened 2018-12-13
Python API: Add hour-based rate limiting policies utilities

aneutron:
##### SUMMARY
Functions to parse hour-based policies for rate limiting, and wrapper around the existing rate limiting method to apply these policies.

##### ISSUE TYPE
 - Feature Pull Request

##### COMPONENT NAME
`Python API`

##### SDS VERSION

```
openio 4.2.6.dev59
```

##### ADDITIONAL INFORMATION

```
Nil.
```

### DIFF
--- a/oio/common/green.py
+++ b/oio/common/green.py
@@ -28,8 +28,6 @@ from eventlet.green.httplib import HTTPConnection, HTTPResponse, _UNKNOWN # noqa
 from eventlet.event import Event # noqa
 from eventlet.queue import Empty, LifoQueue # noqa
 
-from oio.common.exceptions import OioException
-
 eventlet.monkey_patch(os=False)
 
 logging.thread = eventlet.green.thread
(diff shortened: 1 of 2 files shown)


oio/common/green.py
```diff
@@ -77,6 +78,158 @@ def ratelimit(run_time, max_rate, increment=1, rate_buffer=5):
...
+        policy = ratelimit_policy_from_string(policy)
+    ratelimit_validate_policy(policy)
+
+    def _ratelimiter(run_time, *args, **kwargs):
+        """
+        The ratelimit wrapper that takes into account the custom policy, and
+        ignores all the other parameters other than run_time
+        :param run_time: The last time the operation was executed in seconds.
+        """
+        curr_date = datetime.today()
```
AymericDu: To not call twice in a row `time.time()`, we could store the value and transmit it to `ratelimit`.
```suggestion
        curr_time = time.time()
        curr_date = datetime.fromtimestamp(curr_time)
```

fvennetier: Thanks for the suggestion.

PR #1636, opened 2018-12-21
sqliterepo: Clean stale DB dumps

jfsmig:
##### SUMMARY
Clean stale DB dumps at the startup of any sqliterepo-based service, when it owns the lock on the base directory.

##### ISSUE TYPE
Bugfix

##### COMPONENT NAME
`sqliterepo`

##### SDS VERSION
`openio 4.2.6`

### DIFF
--- a/sqliterepo/repository.c
+++ b/sqliterepo/repository.c
@@ -369,11 +369,19 @@ sqlx_repository_initial_cleanup(sqlx_repository_t *repo)
 
 	g_assert_nonnull(repo);
 
-	g_snprintf(pattern, sizeof(pattern), "%s/tmp/dump.sqlite3.*",
-			repo->basedir);
+	/* There are 2 patterns to be managed */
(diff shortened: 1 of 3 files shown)


sqliterepo/repository.c
```diff
@@ -359,6 +360,46 @@ sqlx_repository_init(const gchar *vol, const struct sqlx_repo_config_s *cfg,
...
+void
+sqlx_repository_initial_cleanup(sqlx_repository_t *repo)
+{
+	gchar pattern[PATH_MAX] = {0};
+	glob_t buf = {0};
+	int rc;
+
+	g_assert_nonnull(repo);
+
+	g_snprintf(pattern, sizeof(pattern), "%s/tmp/dump.sqlite3.*",
```
fvennetier: We should also check for `%s/tmp/restore.sqlite3.*`. See https://github.com/open-io/oio-sds/blob/4.2.x/sqliterepo/repository.c#L1584.

PR #1639, opened 2019-01-02
blob-rebuilder: use random delays to avoid request bursts

murlock:
##### SUMMARY

Default ratelimit tries to consume all events at start of
every second: when launching a lot of threads/instance, it
creates burst effects on cluster.

This commit introduce an option to add random wait after each chunk
rebuild, that take an argument to perform a [0, arg[ ms wait.

##### ISSUE TYPE
 - Feature Pull Request

##### COMPONENT NAME
oio-blob-rebuilder

##### SDS VERSION
```
4.2.x
```

### DIFF
--- a/bin/oio-blob-rebuilder
+++ b/bin/oio-blob-rebuilder
@@ -58,6 +58,8 @@ tube for broken chunks events.
                         help="Number of workers (1)")
     parser.add_argument('--chunks-per-second', type=int,
                         help="Max chunks per second per worker (30)")
+    parser.add_argument("--random-wait", type=int,
+                        help="Random wait (in μs)")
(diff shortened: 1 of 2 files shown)


bin/oio-blob-rebuilder
```diff
@@ -58,6 +58,8 @@ tube for broken chunks events.
                         help="Number of workers (1)")
     parser.add_argument('--chunks-per-second', type=int,
                         help="Max chunks per second per worker (30)")
+    parser.add_argument("--random-wait", type=int,
+                        help="Random wait (in ms)")
```
fvennetier: For a better UX, I would use a floating point number of seconds (more pythonic), or an integer number of microseconds (as we use in most of our C programs).

PR #1646, opened 2019-01-10
Update the sample of oio-blob-indexer configuration

AymericDu:
##### SUMMARY

Update the sample of oio-blob-indexer configuration

##### ISSUE TYPE

 - Docs Pull Request

##### COMPONENT NAME

- `oio-blob-indexer`

##### SDS VERSION

```
openio 4.2.7.dev19
```

### DIFF
--- a/etc/blob-indexer.conf-sample
+++ b/etc/blob-indexer.conf-sample
@@ -1,16 +1,33 @@
 [blob-indexer]
+
+# Namespace name
 namespace = NS
+
+# Run daemon as user
 user = openio
-volume = /var/lib/oio/sds/vol1/NS/rawx-1/
-interval = 300
-report_interval = 5
-chunks_per_second = 30
-autocreate = true
+
+# Logging configuration
 log_level = INFO
 log_facility = LOG_LOCAL0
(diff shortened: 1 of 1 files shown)


etc/blob-indexer.conf-sample
```diff
@@ -1,16 +1,33 @@
...
 namespace = NS
+
+# Run daemon as user
 user = openio
+
+# Logging configuration
+log_level = INFO
+log_facility = LOG_LOCAL0
+log_address = /dev/log
+syslog_prefix = OIO,OPENIO,blob-index,1
```
fvennetier: If the namespace is "NS", the syslog prefix should be "OIO,OPENIO,blob-index,1".

etc/blob-indexer.conf-sample
```diff
@@ -1,16 +1,33 @@
...
+
+# Volume to index
 volume = /var/lib/oio/sds/vol1/NS/rawx-1/
+
+# Interval between to passes (in seconds)
```
fvennetier: Interval between ~to~ passes (in seconds)

PR #1649, opened 2019-01-15
core: Allow to send syslog via UDP port 514

jfsmig:
##### SUMMARY
Allows a UDP/IP address (IP:PORT) on the CLI of the services written in C

##### ISSUE TYPE
`enhancement`

##### COMPONENT NAME
`core`

##### SDS VERSION
`openio 4.2.7.dev26`


core/log.c
```diff
@@ -148,32 +148,26 @@ glvl_allowed(register GLogLevelFlags lvl)
...
-	for (gchar c; (c=*s) ; s++) {
-		if (invalid[(guint8)c])
-			*s = ' ';
+void __attribute__ ((constructor))
+_constructor_map_valid_ascii (void)
+{
+	/* 0 remains 0, and the array is already zero'ed */
+	for (int i=1; i<256 ;i++) {
+		map_valid_ascii[i] = (g_ascii_isspace(i) || !g_ascii_isprint(i))
+			? (guint8)' ' : (guint8)i;
```
fvennetier: New lines are converted to spaces. Do we still want that?

jfsmig: I do, but I am not the majority, even in terms of weight.

PR #1650, opened 2019-01-15
tools: Deduplicate code from bootstrap scripts

jfsmig:
##### SUMMARY
Deduplicate code from tools related to the ZK administration. It reuse code in the "oio.zookeeper" module (renamed oio.zk) and now provides a set of `openio` sub commands.

```
# openio zk -h  
Command "zk" matches:
  zk armageddon -> the old zk-reset.py, used by oio-reset.sh
  zk bootstrap -> the old zk-bootstrap.py
  zk reset -> the old oio-election-reset.py
  zk smudge -> the old oio-election-smudge.py
  zk stat -> a mix of oio-election-stat.py and oio-election-dump.py
```

##### ISSUE TYPE
`enhancement`

##### COMPONENT NAME
`tools`

##### SDS VERSION
`openio 4.2.7.dev25`


tools/oio-election-reset.py
```diff
@@ -30,6 +30,25 @@
...
+    def close(self):
+        zookeeper.close(self._zh)
+        self._zh = None
+
+
+def get_connected_handles(cnxstr):
+    zookeeper.set_debug_level(zookeeper.LOG_LEVEL_INFO)
+    for shard in cnxstr.split(";"):
+        zh = zookeeper.init(shard)
+    yield ZkHandle(zh)
```
fvennetier: Looks like this statement should be in the loop.

jfsmig: In facts, the whole has not been deduplicated at all.

PR #1654, opened 2019-01-29
Fix Lifecycle XML parsing when a namespace is provided

fvennetier:
##### SUMMARY
When the Lifecycle configuration is set through swift3, an namespace is set on the XML document. Some adaptations had to be made for the tags to be matched to the expected ones.

##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT NAME
- Python API

##### SDS VERSION

```
openio 4.3.2.dev1
```


oio/container/lifecycle.py
```diff
@@ -244,21 +248,22 @@ def from_element(cls, rule_elt, **kwargs):
...
         try:
             filter_ = LifecycleRuleFilter.from_element(
-                rule_elt.findall('Filter')[-1])
+                rule_elt.findall('Filter', nsmap)[-1])
         except IndexError:
             raise ValueError("Missing 'Filter' element")
 
         try:
-            status = rule_elt.findall('Status')[-1].text
+            status = rule_elt.findall('Status', rule_elt.nsmap)[-1].text
```
AymericDu: We should use `nsmap` (probably an oversight).

cdelgehier:
```
E: Failed to fetch http://mirror.openio.io/pub/repo/openio/sds/18.04/ubuntu/xenial/libzookeeper-java_3.4.13-1oio_all.deb  Could not connect to mirror.openio.io:80 (37.59.36.166), connection timed out
``` true story :s

PR #1658, opened 2019-01-31
tools: Improve the check of copyright mentions

jfsmig:
##### SUMMARY
False negatives happened on empty and removed files. The current PR fixes that. Also produces a lighter debug output.

##### ISSUE TYPE
Between `enhancement` and `fix`

##### COMPONENT NAME
`ci`

##### SDS VERSION
`openio 4.3.2.dev20`


jfsmig:
Let's do this!

PR #1664, opened 2019-02-08
C client: Avoid odd config from the app.

jfsmig:
##### SUMMARY
Avoid odd values from the application, when a chunk_size has been forced. The PR Adds 2 config variable to trigger the check:
* `core.chunk_size.minimum` with a default at **10MB** (set to 0 to disable)
* `core.chunk_size.maximum` with a default of **0** (disabled)

##### ISSUE TYPE
`bug`

##### COMPONENT NAME
`core`, `c sdk`

##### SDS VERSION
`openio 4.2.8.dev26`


conf.json
```diff
@@ -986,6 +987,16 @@
 				"def": "true",
 				"descr": "Should the client adapt metachunk size to EC policy parameters? Letting this on will make bigger metachunks, but chunks on storage will stay at normal chunk size. Disabling this option allows clients to do write alignment." },
 
+			{ "type": "int64", "name": "oio_chunk_size_minimum",
+				"key": "core.chunk_size.minimum",
```
fvennetier: Cosmetic, minor: other variable names use "min" and "max" words instead of "minimum" and "maximum". For example `client.errors_cache.max` and `sqliterepo.repo.fd_min_active`.

jfsmig: I agree. Let's have uniform names and aliases to manage existing productions.

PR #1666, opened 2019-02-11
Improve the `directory bootstrap` fairness on multi-DC

jfsmig:
##### SUMMARY
* Brings a fair distribution of the meta1 prefixes over the multiple sites
* At most one replica of any prefix is on a single site
* Requirement that locations have max 4 tokens, with the site coming first.

##### ISSUE TYPE
`enhancement`

##### COMPONENT NAME
`cli`

##### SDS VERSION
`openio 4.3.2.dev3`

### DIFF
--- a/oio/cli/directory/directory.py
+++ b/oio/cli/directory/directory.py
@@ -106,9 +106,14 @@ class DirectoryInit(DirectoryCmd):
     def get_parser(self, prog_name):
         parser = super(DirectoryInit, self).get_parser(prog_name)
         parser.add_argument(
-            '--level', metavar='<0,1,2,3>', dest='level',
-            choices=(0, 1, 2, 3), default=3,
-            help='Which location token represent the site')
+            '--level', metavar='<LEVEL>', dest='level',
+            choices=('site', 'rack', 'host', 'volume'), default='site',
+            help='Which location level should be perfectly balanced')
+        parser.add_argument(
+            '--degradation', metavar='<DEGRADATION>', dest='degradation',
+            type=int, default=0,
+            help='How many location levels we accept to lose to keep the '
+                 'quorums valid.')
         parser.add_argument(
             '--force',
             action='store_true',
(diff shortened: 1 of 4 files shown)


tests/unit/test_prefixmapping.py
```diff
@@ -18,7 +18,94 @@
...
+        for groups in (prefixes(1),  # 16
+                       prefixes(2),  # 256
+                       prefixes(3),  # 4096
+                       prefixes(4)):  # 64ki
+            groups = list(groups)
+            for replicas in range(2, 8):
+                for sites in range(1, replicas):
+                    srv = self.generate_services(sites*3, nb_sites=sites)
+                    self.assertRaises(Exception, _bootstrap, srv, groups,
+                                      replicas=replicas, level=0, fill_token=2)
```
fvennetier: Here we should expect a specific exception type, to ensure the API does not change silently.

oio/cli/directory/directory.py
```diff
@@ -100,7 +106,9 @@ class DirectoryInit(DirectoryCmd):
     def get_parser(self, prog_name):
         parser = super(DirectoryInit, self).get_parser(prog_name)
         parser.add_argument(
-            '--no-rdir', action='store_true', help='Deprecated')
+            '--level', metavar='<0,1,2,3>', dest='level',
```
fvennetier: In the C language load balancer, the *level* is 0 for the rightmost location token. This can create confusion. Could we talk about _distance_ or _remoteness_ instead?

oio/cli/directory/directory.py
```diff
@@ -35,11 +36,16 @@ class DirectoryCmd(Command):
 
     def get_parser(self, prog_name):
         parser = super(DirectoryCmd, self).get_parser(prog_name)
-        parser.add_argument('--replicas', metavar='<N>', dest='replicas',
-                            type=int, default=3,
-                            help='Set the number of replicas (3 by default)')
-        parser.add_argument('--min-dist', type=int, default=1,
-                            help="Minimum distance between replicas")
+        parser.add_argument(
+            '--no-rdir', action='store_true', help='Deprecated')
```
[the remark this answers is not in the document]
jfsmig: Yes in 4.4.x, it "breaks" few but existing usages of that no-op option.

oio/cli/directory/directory.py
jfsmig: Now more precise, we expect a self-explanatory token ("site", "rack", "host","volume").

oio/cli/directory/directory.py
```diff
@@ -100,7 +106,15 @@ class DirectoryInit(DirectoryCmd):
     def get_parser(self, prog_name):
         parser = super(DirectoryInit, self).get_parser(prog_name)
         parser.add_argument(
-            '--no-rdir', action='store_true', help='Deprecated')
+            '--level', metavar='<LEVEL>', dest='level',
+            choices=('site', 'rack', 'host', 'volume'), default='volume',
```
AymericDu: All customers don't use this format.
We should give an example of location.

PR #1672, opened 2019-02-14
Workaround and fix flaky tests

fvennetier:
##### SUMMARY
Use nosetests' "flaky" module to rerun flaky tests.

##### ISSUE TYPE
 - Feature Pull Request

##### COMPONENT NAME
- tests

##### SDS VERSION
```
openio 4.3.2.dev54
```

### DIFF
--- a/tests/functional/container/test_container.py
+++ b/tests/functional/container/test_container.py
@@ -676,9 +676,8 @@ class TestMeta2Contents(BaseTestCase):
         # Extract one chunk from the list
         chunks.pop()
         if len(chunks) < 1:
-            self.skip(
+            self.skipTest(
                 'Must run with a storage policy requiring more than 1 chunk')
-            return
 
         # Do the spare request, specify that we already know some chunks
         resp = self.request('POST', self.url_content('spare'), params=params,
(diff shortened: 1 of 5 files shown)


fvennetier:
> Is there a possibility to tune `flaky` to make it "retry but not too often"?

By default, it will retry just once.

But it is possible to make it run the test several times, and accept the result after several successes. See https://github.com/box/flaky#marking-tests-flaky for configuration possibilities.

tests/functional/container/test_container.py
```diff
@@ -699,28 +699,42 @@ def test_spare_with_one_broken(self):
...
-        self.assertEqual(1, len(spare_data['chunks']))
-        self.assertEqual(1, len(spare_data['properties']))
-        broken_netloc = broken['url'].split('/')[2]
-        spare_netloc = spare_data['chunks'][0]['id'].split('/')[2]
-        self.assertNotEqual(broken_netloc, spare_netloc)
+        self.assertEqual(count_broken, len(spare_data['chunks']))
+        self.assertEqual(count_broken, len(spare_data['properties']))
+        broken_netlocs = {b['url'].split('/')[2] for b in broken}
+        spare_netlocs = {s['id'].split('/')[2] for s in spare_data['chunks']}
+        self.assertNotEqual(broken_netlocs, spare_netlocs)
```
fvennetier: This test is wrong: we have to check there are no common elements in both sets.
Do not merge before I fix this!

tests/functional/container/test_container.py
```diff
@@ -699,28 +699,43 @@ def test_spare_with_one_broken(self):
         obj_meta = self.json_loads(resp.data)
         # Get a list of chunks for future spare request
         chunks = obj_meta['chunks']
-        if len(self.conf['services']['rawx']) < len(chunks) + 1:
+        if len(chunks) < count_broken + 1:
+            self.skip(
```
murlock: Minor: `self.skip` is deprecated, we should use `self.skipTest`.
and return is useless as `skip` or `skipTest` raise a `skipException`.

PR #1673, opened 2019-02-14
core/lb: allow disabling distance check optimization

fvennetier:
##### SUMMARY

Introduce `core.lb.allow_distance_bypass` variable that disables some load balancing optimizations that may lead to unbalanced situations. Also add new tests.

##### ISSUE TYPE
 - Bugfix Pull Request
 - Feature Pull Request

##### COMPONENT NAME
- LB

##### SDS VERSION
```

```


core/lb.c
```diff
@@ -888,10 +888,11 @@ _local_slot__poll(struct oio_lb_slot_s *slot, const guint16 distance,
...
 	 * we can disable the distance checks, and rely only on the "popularity"
 	 * mechanism.
-	 * XXX: here we compare the overall number of targets to the number
+	 * FIXME: here we compare the overall number of targets to the number
 	 * of different locations IN THE CURRENT SLOT. We bet that in most
 	 * pools there will be only one targetted slot. */
-	if (!reversed && ctx->check_distance && distance > 1) {
+	if (oio_lb_allow_distance_bypass &&
+			!reversed && ctx->check_distance && distance > 1) {
 		guint16 level = distance - 1;
```
jkasarherou: Why immediately degrade distance search??

fvennetier: As said in the comment a few lines above, it is possible in certain situations to detect early that we won't be able to satisfy the current distance constraint.

PR #1678, opened 2019-02-21
proxy: Decache stale service locations more often

jfsmig:
##### SUMMARY
Also decache invalid location entries if clues are present in "group" requests, such the election debug / ping / leave. The PR also fixes a minor memory leak (GError upon a decache) in the proxy.

##### ISSUE TYPE
`bugfix`

##### COMPONENT NAME
`proxy`

##### SDS VERSION
`openio 4.2.8.dev33`

### DIFF
--- a/proxy/common.c
+++ b/proxy/common.c
@@ -63,7 +63,7 @@ service_invalidate (gconstpointer k)
 		GRID_DEBUG("invalid at %lu %s", oio_ext_monotonic_seconds(), (const char*)k);
 }
 
-gboolean
+static gboolean
 service_is_slave (const char *obj, const char *master)
 {
 	gboolean rc;
@@ -73,7 +73,7 @@ service_is_slave (const char *obj, const char *master)
 	return rc;
 }
 
-gboolean
+static gboolean
 service_is_master (const char *obj, const char *master)
 {
 	gboolean rc;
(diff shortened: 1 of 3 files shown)


proxy/common.c
```diff
@@ -372,16 +427,16 @@ gridd_request_replicated (struct req_args_s *args, struct client_ctx_s *ctx,
...
+					flag_prefer_master_for_write)) {
 			const char *actual = client ? gridd_client_url(client) : NULL;
-			// if (actual && 0 != strcmp(actual, _id)) {
 			if (actual && 0 != strcmp(actual, url)) {
-				gchar *k = g_strdup(election_key);
-				gchar *v = g_strdup(actual);
-				GRID_TRACE("MASTER %s %s", v, k);
-				MASTER_WRITE(lru_tree_insert(srv_master, k, v));
+				// TODO(jfs): maybe save the ID instead (or remove this mark)
+				service_learn_master(election_key, actual);
```
fvennetier: If `actual` is "the service that responded without error", and `flag_prefer_slave_for_read` is set, then we are filling the cache with the wrong service.

jfsmig: `actual` is the service we have been redirected to. The only redirection case is the need for a master, the retry with another peer is a piece of logic not managed by the gridd_client API (but by the proxy). So we could save any `actual` that is different from the initial target.

The check on the flags is just here to save a few cycles if possible: if you never use any preference, then you don't need to keep cached information. If the management of preferences becomes request-dependant, the check on the flags must even be removed.

PR #1680, opened 2019-02-21
cmake,ci: Restore the SDK_ONLY option

jfsmig:
##### SUMMARY
There are reports of users expecting the SDK_ONLY option of cmake to be functional.
This is now fixed.

##### ISSUE TYPE
`enhancement`

##### COMPONENT NAME
`cmake`, `ci`

##### SDS VERSION
`openio 4.3.3.dev11`

### DIFF
--- a/tools/oio-travis-tests.sh
+++ b/tools/oio-travis-tests.sh
@@ -270,6 +270,12 @@ if is_running_test_suite "variables" ; then
 	tox -e variables
 fi
 
+if is_running_test_suite "sdk" ; then
+	echo -e "\n### SDK tests"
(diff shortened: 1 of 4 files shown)


tools/oio-travis-tests.sh
```diff
@@ -270,6 +270,12 @@ if is_running_test_suite "variables" ; then
 	tox -e variables
 fi
 
+if is_running_test_suite "sdk" ; then
```
murlock: IMO this is never called as we have a `exit 0` just after `make test` with `SDK` mode in `.travis.yml`

PR #1682, opened 2019-02-22
Permit to only list buckets

kamel-rahim:
##### SUMMARY

Add a parameter to the container listing to get only buckets

##### ISSUE TYPE

 - Feature Pull Request

##### COMPONENT NAME

* CLI
* Python API
##### SDS VERSION

```
openio 4.4.0.0b1.dev15
```

### DIFF
--- a/oio/account/backend.py
+++ b/oio/account/backend.py
@@ -16,6 +16,7 @@
 
 from time import time
 
+import re
 import redis
 import redis.sentinel
 from werkzeug.exceptions import NotFound, Conflict, BadRequest
@@ -188,6 +189,21 @@ class AccountBackend(RedisConn):
             redis.call('DEL', KEYS[3] .. container);
         end;
         """
+
+    # This regex comes from https://stackoverflow.com/a/50484916
+    #
+    # The first group looks ahead to ensure that the match
+    # is between 3 and 63 characters long.
+    #
+    # The next group (?!^(\d+\.)+\d+$) looks ahead to forbid matching
+    # bucket names that look like IP addresses.
+    #
+    # <The last group matches zero or more labels followed by a dot *
+    buckets_pattern = re.compile(
+        r"""(?=^.{3,63}$)   # first group
+        (?!^(\d+\.)+\d+$) # second
+        (^(([a-z0-9]|[a-z0-9][a-z0-9\-]*[a-z0-9])\.)* #third
+        ([a-z0-9]|[a-z0-9][a-z0-9\-]*[a-z0-9])$)""", re.X)
 
     def __init__(self, conf, connection=None):
         self.conf = conf
@@ -362,8 +378,11 @@ class AccountBackend(RedisConn):
 
         return name
 
+    def _should_be_listed(self, c_id, s3_buckets_only):
+        return not s3_buckets_only or self.buckets_pattern.match(c_id)
+
     def _raw_listing(self, account_id, limit, marker, end_marker, delimiter,
-                     prefix):
+                     prefix, s3_buckets_only=False):
         """Fetch tuple list of containers matching options.
            Tuple is [(container|prefix),
                      0 *reserved for objects*,
@@ -393,15 +412,20 @@ class AccountBackend(RedisConn):
                              for cid in container_ids]
 
             if prefix is None:
-                containers = [[c_id, 0, 0, 0, 0] for c_id in container_ids]
+                containers = [[c_id, 0, 0, 0, 0] for c_id in container_ids
+                              if self._should_be_listed(c_id, s3_buckets_only)]
                 return containers
             if not delimiter:
                 if not prefix:
-                    containers = [[c_id, 0, 0, 0, 0] for c_id in container_ids]
+                    containers = [[c_id, 0, 0, 0, 0] for c_id in container_ids
+                                  if self._should_be_listed(c_id,
+                                                            s3_buckets_only)]
                     return containers
                 else:
                     containers = [[c_id, 0, 0, 0, 0] for c_id in container_ids
-                                  if c_id.startswith(prefix)]
+                                  if c_id.startswith(prefix) and
+                                  self._should_be_listed(c_id,
+                                                         s3_buckets_only)]
                     return containers
 
             count = 0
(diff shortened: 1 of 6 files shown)


oio/account/backend.py
```diff
@@ -418,16 +418,20 @@ def _raw_listing(self, account_id, limit, marker, end_marker, delimiter,
                     if dir_name != orig_marker:
                         results.append([dir_name, 0, 0, 1, 0])
                     break
-                results.append([container_id, 0, 0, 0, 0])
+                if not buckets or (buckets
+                                   and "+segments" not in container_id
+                                   and "%2F" not in container_id):
```
AymericDu: We could simplify the condition:
```
not buckets or ('+segments' not in container_id and '%2F' not in container_id)
```

tests/functional/api/test_objectstorage.py
```diff
@@ -687,6 +687,18 @@ def __call__(self, *args, **kwargs):
...
 
+    def test_bucket_list(self):
+        account = random_str(32)
+        name_with_path = "%2F".join([random_str(32), random_str(32)])
+        name_with_segments = random_str(32) + "+segments"
+        name = random_str(32)
+        self.api.container_create(account, name)
+        self.api.container_create(account, name_with_path)
+        self.api.container_create(account, name_with_segments)
+        buckets = self.api.container_list(account)
```
AymericDu: The `buckets=True` parameter is missing.

tests/unit/api/test_objectstorage.py
```diff
@@ -88,7 +88,8 @@ def test_container_list(self):
             self.account, limit=limit, marker=marker, prefix=prefix,
             delimiter=delimiter, end_marker=end_marker, headers=self.headers)
         params = {"id": self.account, "prefix": prefix, "delimiter": delimiter,
-                  "marker": marker, "end_marker": end_marker, "limit": limit}
+                  "marker": marker, "end_marker": end_marker, "limit": limit,
+                  "buckets": False}
```
AymericDu: It's not mandatory.

oio/account/server.py
```diff
@@ -365,10 +365,11 @@ def on_account_containers(self, req):
         prefix = req.args.get('prefix', '')
         limit = int(req.args.get('limit', '1000'))
         delimiter = req.args.get('delimiter', '')
+        buckets = bool(req.args.get('buckets', False))
```
AymericDu: We should use `true_value` or `boolean_value` (https://github.com/open-io/oio-sds/blob/master/oio/common/easy_value.py#L44)

oio/account/backend.py
murlock: Technically, bucket name should respect https://docs.aws.amazon.com/AmazonS3/latest/dev/BucketRestrictions.html#bucketnamingrules

tests/unit/api/test_objectstorage.py
kamel-rahim: I've got this when I remove it
```
AssertionError: Expected call: mock('GET', 'http://fake_endpoint/v1.0/account/containers', headers={'X-oio-req-id': '18ABE4149AABF93282C50E57DFAA79DF', 'x-oio-req-id': 'XVhTd2L3V4AE1tLR9nOjKmhCpm36k8pf'}, params={'end_marker': 'MYLakA0MRfip1U7eQ0FkTJGOxHyWJe8W', 'prefix': 'lkJy9UmjckCySCif6DRGQeoafF1Vq5rz', 'delimiter': 'Wn9N5AX84JZBzvcROd1T6s8W9etA8ZiS', 'limit': 653, 'marker': 'GbRQdxwGAAmWYmcdqAVen0F5uLi596EA', 'id': 'test'})
```
```
Actual call: mock('GET', 'http://fake_endpoint/v1.0/account/containers', headers={'X-oio-req-id': '18ABE4149AABF93282C50E57DFAA79DF', 'x-oio-req-id': 'XVhTd2L3V4AE1tLR9nOjKmhCpm36k8pf'}, params={'end_marker': 'MYLakA0MRfip1U7eQ0FkTJGOxHyWJe8W', 'buckets': False, 'prefix': 'lkJy9UmjckCySCif6DRGQeoafF1Vq5rz', 'delimiter': 'Wn9N5AX84JZBzvcROd1T6s8W9etA8ZiS', 'limit': 653, 'marker': 'GbRQdxwGAAmWYmcdqAVen0F5uLi596EA', 'id': 'test'})
```

oio/account/client.py
```diff
@@ -148,7 +148,7 @@ def account_update(self, account, metadata, to_delete, **kwargs):
 
     def container_list(self, account, limit=None, marker=None,
                        end_marker=None, prefix=None, delimiter=None,
-                       **kwargs):
+                       buckets=False, **kwargs):
```
fvennetier: Since it is not documented, this parameter lets me think buckets **won't** be listed by default.
I suggest:
- renaming this parameter so it tells what it does;
- documenting it.

oio/account/backend.py
AymericDu: We could use a regex: https://stackoverflow.com/questions/50480924/regex-for-s3-bucket-name

kamel-rahim: Are you sure to use this regex `(?=^.{3,63}$)(?!^(\d+\.)+\d+$)(^(([a-z0-9]|[a-z0-9][a-z0-9\-]*[a-z0-9])\.)*([a-z0-9]|[a-z0-9][a-z0-9\-]*[a-z0-9])$)` ?

oio/account/client.py
kamel-rahim: I've add a small doc on it.
But there is not many names that fit this options. Maybe `listing_buckets`.

fvennetier: `s3_buckets_only`? `buckets_only`?

oio/account/backend.py
AymericDu: Yes, if you explain each part.
This regex is used to check all cases. But your conditions do not check the following 2 cases:
- Bucket names must be a series of one or more labels. Adjacent labels are separated by a single period (.). Bucket names can contain lowercase letters, numbers, and hyphens. Each label must start and end with a lowercase letter or a number.
- Bucket names must not be formatted as an IP address (for example, 192.168.5.4).

PR #1688, opened 2019-02-26
oio-blob-converter: restore fullpath if grid.content.path or  oio:HASH are missing

kamel-rahim:
##### SUMMARY

Recreate fullpath  when xattrs 'grid.contant.path' and 'oio:$HASH' are also missing

##### ISSUE TYPE

 - Bugfix Pull Request

##### COMPONENT NAME
oio-blob-converter

##### SDS VERSION

```
openio 4.2.8.dev36
```

### DIFF
--- a/oio/blob/converter.py
+++ b/oio/blob/converter.py
@@ -374,7 +374,9 @@ class BlobConverter(object):
 
     def is_fullpath_error(self, err):
         if (isinstance(err, MissingAttribute) and
-                err.attribute.startswith(CHUNK_XATTR_CONTENT_FULLPATH_PREFIX)):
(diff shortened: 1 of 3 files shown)


oio/common/constants.py
```diff
@@ -100,6 +100,7 @@
 }
 
 CHUNK_XATTR_CONTENT_FULLPATH_PREFIX = 'oio.content.fullpath:'
+CHUNK_XATTR_CONTENT_FULLPATH_OLD_PREFIX = 'oio:'
```
AymericDu: This constant already exists: https://github.com/open-io/oio-sds/blob/4.2.x/oio/blob/converter.py#L42

kamel-rahim: Done

PR #1690, opened 2019-02-27
sqliterepo: minor fixes

murlock:
##### SUMMARY

`completion_LISTING` did not check validity of argument before using it.
Also fix a codepath with strrchr could be called on NULL

##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT NAME
sqliterepo

##### SDS VERSION
```
4.4.0.0b1.dev35
```

### DIFF
--- a/sqliterepo/election.c
+++ b/sqliterepo/election.c
@@ -1636,40 +1636,42 @@ static void
 completion_LISTING(int zrc, const struct String_vector *sv, const void *d)
 {
 	if (!d) return;
-	if (!sv) return;
 
 	struct election_member_s *member = (struct election_member_s*) d;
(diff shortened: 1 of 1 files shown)


sqliterepo/election.c
```diff
@@ -1636,6 +1636,7 @@ static void
 completion_LISTING(int zrc, const struct String_vector *sv, const void *d)
 {
 	if (!d) return;
+	if (!sv) return;
```
jfsmig: This would lead to an election lost in its `LISTING` state. It is necessary that completion hooks make the FSM move, even if it is toward the FAILED state.

PR #1691, opened 2019-03-04
Register proxy and beanstalkd into conscience

murlock:
##### SUMMARY
Register `proxy` and `beanstalkd` into `conscience`, it will allow us to launch commands on all cluster.

##### ISSUE TYPE
 - Feature Pull Request

##### COMPONENT NAME
conscience-agent

##### SDS VERSION
```
master
```

### DIFF
--- a/oio/conscience/checker/beanstalk.py
+++ b/oio/conscience/checker/beanstalk.py
@@ -1,0 +1,53 @@
+# Copyright (C) 2019 OpenIO SAS, as part of OpenIO SDS
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Affero General Public License as
+# published by the Free Software Foundation, either version 3 of the
+# License, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+from oio.common import exceptions as exc
+from oio.event.beanstalk import Beanstalk, ConnectionError
+from oio.conscience.checker.base import BaseChecker
+
+
+class BeanstalkChecker(BaseChecker):
+    name = 'beanstalkd'
+
+    def configure(self):
+        for k in ('host', 'port'):
+            if k not in self.checker_conf:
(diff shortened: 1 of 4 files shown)


oio/conscience/checker/beanstalk.py
```diff
@@ -0,0 +1,44 @@
...
+                raise exc.ConfigurationException(
+                    'Missing field "%s" in configuration' % k)
+        print("YYY")
+
+    def check(self):
+        result = False
+        cnx = None
+        try:
+            cnx = Beanstalk(self.checker_conf['host'], self.checker_conf['port'])
+            cnx.stats_tube(self.checker_conf['tube'])
```
fvennetier: I think we only need to check the state of the service, not a specific tube. Calling `stats` would be better (must be implemented in `beanstalk.py`).

tools/oio-bootstrap.py
```diff
@@ -426,6 +426,32 @@ def save_coverage():
...
+host: ${IP}
+port: ${PORT}
+type: beanstalkd
+location: ${LOC}
+checks:
+    - {type: beanstalk, tube: oio}
+slots:
+    - beanstalk
+stats:
+    - {type: system}
```
fvennetier: Beanstalkd services store data on the file system, thus we should monitor the volume they are using.
```suggestion
    - {type: system}
    - {type: volume, path: ${VOLUME}}
```

tools/oio-bootstrap.py
```diff
@@ -426,6 +426,32 @@ def save_coverage():
...
 
+template_beanstalk_watch = """
+host: ${IP}
+port: ${PORT}
+type: beanstalkd
+location: ${LOC}
+checks:
+    - {type: beanstalk, tube: oio}
+slots:
+    - beanstalk
```
fvennetier: To remove any ambiguity, we should write beanstalk**d** everywhere.

murlock: :+1:

tools/oio-bootstrap.py
murlock: Done

tools/oio-bootstrap.py
```diff
@@ -426,6 +426,32 @@ def save_coverage():
...
+host: ${IP}
+port: ${PORT}
+type: beanstalkd
+location: ${LOC}
+checks:
+    - {type: beanstalkd}
+slots:
+    - beanstalkd
+stats:
+    - {type: system, path: ${VOLUME}}
```
fvennetier: Still not good.
```suggestion
    - {type: system}
    - {type: volume, path: ${VOLUME}}
```

PR #1699, opened 2019-03-13
sqliterepo: Bound the memory usage

jfsmig:
##### SUMMARY
First optimistic memory usage limitation.

When a base is closed and returns to the cache, if the memory limit has been reached the oldest base in cache is expired. The sqlite3 pages are returned to the sqlite3 memory pool and neither to the glibc nor to the kernel). Thus they become available for existing or new bases but you won't observe the RSS usage to shrink.

Adds a new central variable `sqliterepo.rss.max` that is set to 0 by default, and 0 means the maximum is deduced for `getrlimit(RLIMIT_DATA)`.

##### ISSUE TYPE
`enhancement`

##### COMPONENT NAME
`sqliterepo`

##### SDS VERSION
`openio 4.4.0.0b0.dev67`


sqlx/sqlx_service.c
```diff
@@ -241,6 +241,21 @@ _configure_with_arguments(struct sqlx_service_s *ss, int argc, char **argv)
...
 }
 
+static void
+_patch_configuration_maxrss(void)
+{
+	if (sqliterepo_max_rss <= 0) {
+		struct rlimit rl = {};
+		int rc = getrlimit(RLIMIT_DATA, &rl);
+		if (rc == 0) {
+			gint64 maxrss = rl.rlim_cur;
```
murlock: weird indentation

sqlx/sqlx_service.c
```diff
@@ -241,6 +241,21 @@ _configure_with_arguments(struct sqlx_service_s *ss, int argc, char **argv)
...
+		    if (maxrss <= 0)
+				maxrss = G_MAXINT64;
+			sqliterepo_max_rss = maxrss;
```
murlock: we should emit a log with detected `sqliterepo_max_rss`

sqliterepo/cache.c
```diff
@@ -105,6 +106,30 @@ struct sqlx_cache_s
...
+{
+#if HAVE_MALLINFO
+	/* JFS: check on the glibc level. It would not detect a memory exhaust
+	 * if another library would be in use (e.g. jemalloc, dietlib, etc). But
+	 * it causes no syscall and currently matches our usage on the field
+	 * (where the glibc is used). */
+	if (sqliterepo_max_rss > 0) {
+		struct mallinfo mi = mallinfo();
+		gint64 total = 0;
+		total += mi.arena;
```
murlock: `arena` is in bytes and other members (`ordblks`, ...) use block as unit

sqliterepo/cache.c
```diff
@@ -105,6 +106,30 @@ struct sqlx_cache_s
...
+		total += mi.ordblks;
+		total += mi.smblks;
+		total += mi.hblkhd;
+		GRID_WARN("RAM [MiB] used %" G_GINT64_FORMAT" max %" G_GINT64_FORMAT,
+				total / (1024 * 1024),
+				sqliterepo_max_rss / (1024 * 1024));
```
fvennetier: I suggest using DEBUG log level, at least when `total < sqliterepo_max_rss`.

conf.json
```diff
@@ -350,6 +350,11 @@
 		"code": "sqliterepo_variables.c",
 		"header": "sqliterepo_variables.h",
 		"variables": [
+			{ "type": "int64", "name": "sqliterepo_max_rss",
+				"key": "sqliterepo.rss.max",
+				"descr": "Memory size ceiling we try to honor. The check is performed while opening databases. Set to 0 to autodetect the max RSS from the resource limits (see `man getrlimit` for more information). The autodetection considers the environment and is not aware of all the processes that could share that environment.",
+				"def": "0", "min": "0", "max": "max"},
```
fvennetier: The check is performed while **closing** databases.

sqliterepo/cache.c
```diff
@@ -105,6 +106,30 @@ struct sqlx_cache_s
...
+	 * if another library would be in use (e.g. jemalloc, dietlib, etc). But
+	 * it causes no syscall and currently matches our usage on the field
+	 * (where the glibc is used). */
+	if (sqliterepo_max_rss > 0) {
+		struct mallinfo mi = mallinfo();
+		gint64 total = 0;
+		total += mi.arena;
+		total += mi.ordblks;
+		total += mi.smblks;
+		total += mi.hblkhd;
```
fvennetier: `arena` and `hblkhd` should be sufficient ("The total amount of memory allocated by means other than mmap" and "The number of bytes in blocks currently allocated using mmap").

sqlx/sqlx_service.c
jfsmig: Emitting logs in signal handlers is not an excellent idea (despite the fact we already do).
The glib log handler might requires locking, and the signal handler might be fired while a log was already being emitted, leading to a deadlock.

jfsmig:
##### Thought of the day
Prior to this PR we detected the memory leaks with a RSS size growing monotonically.
If we bound the RSS usage with an action on sqliterepo, it will become more difficult to detect memory leaks.

PR #1704, opened 2019-03-19
CLI: implement 'openio-admin <item_type> locate' commands

fvennetier:
##### SUMMARY
- Create mixin classes for container and object commands.
- Implement 'openio-admin <item_type> locate' commands.

##### ISSUE TYPE
 - Feature Pull Request

##### COMPONENT NAME
- CLI

##### SDS VERSION
```
openio 4.4.0.0b2.dev14
```


oio/cli/admin/common.py
```diff
@@ -58,15 +59,33 @@ def patch_parser(self, parser):
...
+            dest='is_cid',
+            help="Interpret <container_name> as a container ID",
+        )
+
+    # TODO(FVE): merge with oio.cli.object.object.ContainerCommandMixin
+    def take_action(self, parsed_args):
+        if not parsed_args.container and not parsed_args.auto:
+            from argparse import ArgumentError
+            raise ArgumentError(parsed_args.container,
+                                "Missing value for container or --auto")
```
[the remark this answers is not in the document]
fvennetier: See the comment below: there is no possibility of specifying an object name without filling the `container` variable. Thus when `--auto` is specified, the `container` variable is interpreted as the name of an object.

oio/cli/admin/item_locate.py
```diff
@@ -0,0 +1,393 @@
...
+
+DUMMY_SERVICE = {'addr': None, 'score': 0, 'tags': {}}
+
+
+class ItemLocateCommand(lister.Lister):
+    """
+    A command to display which service is in charge of hosting an item.
+    """
+
+    columns = ('Type', 'Item', 'Service Id', 'Addr', 'Location', 'Status')
```
AymericDu: We could add `Volume`.

AymericDu: We could replace `Status` with 2 columns (`Up` and `Score`)

fvennetier: What would we display in the `Volume` column?

I don't like the idea of splitting `Status`: currently it displays either the status (in the form of a boolean and a score) or the error that occurred during the check. Where would we display the error then?

PR #1708, opened 2019-03-20
Python API: use reqid instead of req_id

fvennetier:
##### SUMMARY
Try to be consistent when using abbreviations. We choose reqid over req_id since it was more common in the existing code base, especially in log messages.

##### ISSUE TYPE
 - Docs Pull Request

##### COMPONENT NAME
- Python API

##### SDS VERSION
```
openio 4.4.0.0b2.dev18
```

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -12,3 +12,22 @@ zealous and modify more lines than expected.
 
 The code must comply with [PEP8](https://www.python.org/dev/peps/pep-0008/).
 Running ``tox -e pep8`` will tell what is wrong.
+
+
+## Abbreviations
+
+When using abbreviations, try to be consistent. Here is a list of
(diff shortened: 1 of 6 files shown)


CONTRIBUTING.md
```diff
@@ -12,3 +12,22 @@ zealous and modify more lines than expected.
...
 Running ``tox -e pep8`` will tell what is wrong.
+
+
+## Abbreviations
+
+When using abbreviations, try to be consistent. Here is a list of
+abbreviation commonly used in the existing code base.
+
+- acct: account
+- cid: container ID
```
kamel-rahim: What about chunk id or content id ?

fvennetier: We don't have common abbreviations for these ones. "Content" is almost the same as "object", so we could use `oid` for "object ID".

PR #1718, opened 2019-04-01
Decode JSON only if the content type is 'application/json'

AymericDu:
##### SUMMARY

Decode JSON only if the content type is `application/json`.

##### ISSUE TYPE

 - Bugfix Pull Request

##### COMPONENT NAME

- Python API
- `account`
- `proxy`

##### SDS VERSION

```
openio 4.4.2.dev30
```

### DIFF
--- a/oio/api/base.py
+++ b/oio/api/base.py
@@ -25,7 +25,8 @@ from oio.common import exceptions
 from oio.common.utils import deadline_to_timeout, monotonic_time
 from oio.common.constants import ADMIN_HEADER, \
     TIMEOUT_HEADER, PERFDATA_HEADER, FORCEMASTER_HEADER, \
-    CONNECTION_TIMEOUT, READ_TIMEOUT, REQID_HEADER, STRLEN_REQID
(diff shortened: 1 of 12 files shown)


oio/api/base.py
```diff
@@ -188,11 +191,11 @@ def _direct_request(self, method, url, headers=None, data=None, json=None,
         try:
             resp = pool_manager.request(method, url, **out_kwargs)
             body = resp.data
-            if body:
+            if body and resp.headers.get('Content-Type') == "application/json":
                 try:
                     body = jsonlib.loads(body)
                 except ValueError:
-                    pass
+                    raise OioException("Response body isn't decodable JSON")
```
fvennetier: Adding a warning message with the exception here would help debugging.

oio/api/base.py
```diff
@@ -189,11 +190,14 @@ def _direct_request(self, method, url, headers=None, data=None, json=None,
                 service_perfdata['overall'] = service_perfdata.get(
                     'overall', 0.0) + request_end - request_start
             body = resp.data
-            if body:
+            if body and resp.headers.get('Content-Type') == "application/json":
```
fvennetier: HTTP_CONTENT_TYPE_JSON

PR #1721, opened 2019-04-03
Fix container snapshot feature

murlock:
##### SUMMARY

The snapshot operation did not show the created container in account service:
- the `sys.user.name` and `sys.account` was not updated
- a manual `openio container touch` was not possible due to these erroneous values

Fixes OS-350

##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT NAME
CLI

##### SDS VERSION
```
openio 4.2.9.dev10
```

### DIFF
--- a/oio/api/object_storage.py
+++ b/oio/api/object_storage.py
@@ -439,6 +439,7 @@ class ObjectStorageApi(object):
                     target_beans, copy_beans,
                     dst_account, dst_container,
                     frozen=True, **kwargs)
+            self.container.container_touch(dst_account, dst_container)
         finally:
             self.container.container_enable(account, container, **kwargs)
 
--- a/oio/event/filters/dump.py
+++ b/oio/event/filters/dump.py
@@ -1,0 +1,42 @@
+# Copyright (C) 2019 OpenIO SAS, as part of OpenIO SDS
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Affero General Public License as
+# published by the Free Software Foundation, either version 3 of the
+# License, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+from oio.common.json import json
+from oio.event.filters.base import Filter
+
+
+class DumpFilter(Filter):
+    """
+    Dump events to files in /tmp directory.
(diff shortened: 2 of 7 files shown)


oio/api/object_storage.py
```diff
@@ -411,6 +411,11 @@ def container_snapshot(self, account, container, dst_account,
             self.container.container_freeze(account, container, **kwargs)
             self.container.container_snapshot(
                 account, container, dst_account, dst_container, **kwargs)
+            # TODO(mbo) maybe it should be done in proxy
+            self.container.container_set_properties(
+                dst_account, dst_container,
+                system={'sys.user.name': dst_container,
+                        'sys.account': dst_account})
```
fvennetier: If the snapshot is supposed to be a "real" container, and not just a snapshot, then yes, this should be done by oio-proxy. I thought for one moment that this could have been done even one level below, but since this uses an sqliterepo feature, there is no notion of container or account (`sys.account` and `sys.user.name` are meta2-only entries).

oio/event/filters/dump.py
```diff
@@ -0,0 +1,40 @@
...
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+from oio.common.json import json
+from oio.common.logger import get_logger
+from oio.event.filters.base import Filter
+
+
+class DumpFilter(Filter):
```
fvennetier: This whole file seems unnecessary, especially if it has no documentation. Please document it or use the `logger` filter (which dumps all events to syslog).

tests/functional/api/test_objectstorage.py
```diff
@@ -1087,6 +1087,11 @@ def test_container_snapshot(self):
         # Non existing snapshot should work
         self.api.container_snapshot(self.account, container, self.account,
                                     snapshot)
+        # Check sys.user.name is correct
+        ret = self.api.container_get_properties(self.account, snapshot)
+        self.assertEqual(ret['system']['sys.user.name'], snapshot)
+        self.assertEqual(ret['system']['sys.account'], self.account)
```
fvennetier: The assertions are supposed to be done in the reverse order: *expected* before *actual*. Some assertion methods do not specify it, other (like `assertDictEqual`) do and produce better error messages is the order is respected.

proxy/m2_actions.c
```diff
@@ -1336,59 +1336,73 @@ _m2_container_snapshot(struct req_args_s *args, struct json_object *jargs)
...
-
-	err = _resolve_meta2(args, CLIENT_PREFER_MASTER, _pack, NULL, NULL);
+	err = _resolve_meta2(args, CLIENT_PREFER_MASTER,
+			_pack_snapshot, NULL, NULL);
 	g_free(url);
-	if(err)
-		goto cleanup;
 
-cleanup:
+	/* Give the snapshot its name. */
```
murlock: :+1:

PR #1731, opened 2019-04-11
Fix memory leaks in meta2 and meta2 tests

fvennetier:
##### SUMMARY
Fix small memory leaks in meta2 service and related tests.

##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT NAME
- meta2

##### SDS VERSION
```
openio 4.4.2.dev44
```

### DIFF
--- a/meta2v2/meta2_backend.c
+++ b/meta2v2/meta2_backend.c
@@ -839,6 +839,7 @@ _is_damaged_object(struct meta2_backend_s *m2b, struct sqlx_sqlite3_s *sq3,
 		_bean_cleanl2(beans);
 	}
 
+	namespace_info_free(nsinfo);
 	if (err) {
 		GRID_WARN("Impossible to know if object is damaged (%s): %s",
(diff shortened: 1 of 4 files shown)


meta2v2/meta2_backend.c
```diff
@@ -1075,8 +1076,10 @@ meta2_backend_delete_alias(struct meta2_backend_s *m2b,
 					deleted_bean=deleted_bean->next) {
 				cb(u0, deleted_bean->data);
 			}
+			g_slist_free(deleted_beans);
+		} else {
+			_bean_cleanl2(deleted_beans);
```
kamel-rahim: If I understood it correctly. `_bean_cleanl2` is only use during test where the function `cb` is not set and we don't free the data inside the list, isn't it ?

fvennetier: The callback takes possession of the list elements, and frees them later. When there is no callback, the elements are not freed, thus we must call `_bean_cleanl2` (or equivalent) to free both the list and the elements.

PR #1736, opened 2019-04-15
Allow to dispatch tasks locally and across the platform

AymericDu:
##### SUMMARY

Allow to dispatch locally and across the platform.

##### ISSUE TYPE

 - Feature Pull Request

##### COMPONENT NAME

- Python API
- `oio-blob-rebuilder`

##### SDS VERSION

```
openio 5.0.0.0a1.dev51
```


oio/common/tool.py
```diff
@@ -0,0 +1,538 @@
...
+                beanstalkd = all_available_beanstalkd.get(
+                    local_service['addr'])
+                if beanstalkd is None:
+                    continue
+                if beanstalkd_reply \
+                        and beanstalkd_reply['score'] >= beanstalkd['score']:
+                    continue
+                beanstalkd_reply = beanstalkd
+        except Exception as exc:  # pylint: disable=broad-except
+            self.logger.warning('ERROR when searching for beanstalkd locally')
```
fvennetier: The exception message should be printed somehow.

bin/oio-blob-rebuilder
```diff
@@ -35,58 +34,69 @@ before running this tool. This tool can also keep listening to a beanstalkd
...
-    parser.add_argument('--report-interval', type=int,
-                        help="Report interval in seconds (3600)")
-    parser.add_argument('--workers', type=int,
-                        help="Number of workers (1)")
-    parser.add_argument('--chunks-per-second', type=int,
-                        help="Max chunks per second per worker (30)")
-    parser.add_argument("--random-wait", type=int,
-                        help="Random wait (in microseconds)")
-    parser.add_argument('--allow-same-rawx', action='store_true',
-                        help="Allow rebuilding a chunk on the original rawx")
```
fvennetier: I think we should keep the `--allow-same-rawx` option, and print a warning telling it is now enabled by default.

bin/oio-blob-rebuilder
```diff
@@ -35,58 +34,75 @@ before running this tool. This tool can also keep listening to a beanstalkd
...
+             '(default=%s)' % BlobRebuilder.DEFAULT_ALLOW_SAME_RAWX)
+
+    # distributed
+    parser.add_argument(
+        '--distributed', action='store_true',
+        help='Send broken chunks to beanstalkd tubes '
+             'instead of rebuilding them locally '
+             '(the following options are ignored: '
+             '--dry-run, --workers, --chunks-per-second, '
+             '--delete-faulty-chunks).')
```
kamel-rahim: I think that It would be better that `--dry-run` ignore  `--distributed`.

PR #1744, opened 2019-04-19
CLI: implement get-config and set-config commands

fvennetier:
##### SUMMARY
These commands allow get/set ephemeral configuration parameters on the
service.

##### ISSUE TYPE
 - Feature Pull Request

##### COMPONENT NAME
- CLI
- Python API

##### SDS VERSION
```
openio 4.4.3.dev25
```


fvennetier:
> The service type isn't checked.

I chose to keep the service type in the command line to mimic other commands, but indeed it doesn't matter (except for `proxy` which uses a different mechanism).

PR #1746, opened 2019-04-21
rawx: Performance improvements 

jfsmig:
##### SUMMARY
**TL;DR**: the golang rawx is now  much faster but requires a new dependency on `golang.org/x/sys/unix`

* GET now manages to use the `sendfile()` function
* Larger buffers at upload
* file-descriptor oriented xattr operations
* syscalls parsimony
* post-reply asynchronous notifications with a buffered channel

Known TODO: make the golang rawx multi-tube.

##### ISSUE TYPE
`enhancement`

##### COMPONENT NAME
`rawx`

##### SDS VERSION
`openio 4.4.3.dev6`


fvennetier:
This should be rebased on the _master_ branch since it is a rather big change and requires a new dependency.

PR #1750, opened 2019-04-25
ci: Use a travis pipeline

jfsmig:
##### SUMMARY
* Fail fast on gotchas, then proceed with the long tests.
* Attempt to split the logic of the tests (in `tools/oio-test-*`) and the Travis environment specifics (in `tools/oio-travis-*`), so that the logic should be easy to run on a dev sandbox.

##### ISSUE TYPE
`enhancement`

##### COMPONENT NAME
`ci`

##### SDS VERSION
`openio 5.0.0.0a2.dev1`


.travis.yml
```diff
@@ -52,42 +52,59 @@ install:
...
-  - TEST_SUITE=webhook
-  - TEST_SUITE=slave
-  - TEST_SUITE=cli
-  - TEST_SUITE=worm
-  - TEST_SUITE=build,unit,copyright,variables
-  - TEST_SUITE=mover,with-service-id
+  global:
+  - CMAKE_OPTS="-DENABLE_CODECOVERAGE=on -DCMAKE_INSTALL_PREFIX=/tmp/oio -DLD_LIBDIR=lib -DZK_LIBDIR=/usr/lib -DZK_INCDIR=/usr/include/zookeeper -DAPACHE2_LIBDIR=/usr/lib/apache2 -DAPACHE2_INCDIR=/usr/include/apache2 -DAPACHE2_MODDIR=/tmp/oio/lib/apache2/module"
+  - G_DEBUG=fatal_warnings
+  - G_SLICE=always-malloc
```
fvennetier: `G_DEBUG=fatal_warnings` is fine, but we may accelerate several tests if we stop using `G_SLICE=always-malloc`.

PR #1755, opened 2019-04-30
Improve Admin CLI

AymericDu:
##### SUMMARY

- Limit speed of distributed dispatcher
- Use the same column `Errors` to display the errors
- Fail the command if there is at least one error
- Set the default verbose level to INFO
- Use common mixin

##### ISSUE TYPE

 - Bugfix Pull Request

##### COMPONENT NAME

- Python API
- Admin CLI

##### SDS VERSION

```
openio 5.0.0.0a2.dev20
```

### DIFF
--- a/oio/cli/admin/service_check.py
+++ b/oio/cli/admin/service_check.py
@@ -280,9 +280,13 @@ class RawxCheck(ItemCheckCommand):
     def _format_results(self, checker):
         for res in checker.run():
             if res.target.type == 'chunk':
-                yield repr(res.target)[len('chunk='):], \
-                    res.health,\
-                    res.errors_to_str()
+                if not res.errors:
+                    status = 'OK'
+                else:
+                    status = 'error'
+                    self.success = False
+                yield (repr(res.target)[len('chunk='):],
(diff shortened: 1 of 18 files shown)


oio/cli/admin/service_check.py
```diff
@@ -280,42 +284,35 @@ def formatter_default(self):
...
             reqid = self.app.request_id(self.reqid_prefix)
             chunks = self.app.client_manager.rdir.chunk_fetch(service,
                                                               reqid=reqid)
-            for res in self.check_chunks(service, chunks, checker):
+            for res in self.check_chunks(service, chunks, self.checker):
                 yield res
 
     def check_chunks(self, service, chunks, checker):
-        url = self._ensure_http_prefix(service)
+        url = 'http://' + service
```
murlock: Is it compatible with `Service ID` ?

AymericDu: Yes, it's compatible with `Service ID`.
https://github.com/open-io/oio-sds/blob/master/oio/blob/cache.py#L53

PR #1757, opened 2019-05-03
ci: Introduce Docker-based tests

jfsmig:
##### SUMMARY
* Provide a CircleCI configuration, currently only running the fail-fast and unit tests.
* Currently builds on a Ubuntu 18.04 Bionic
* Move the CI scripts into the `ci/` subdirectory.

TODO: rely on a custom image instead of a base ubuntu

##### ISSUE TYPE
`enhancement`

##### COMPONENT NAME
`ci`

##### SDS VERSION
`openio 5.0.0.0a2.dev20`


.circleci/config.yml
```diff
@@ -0,0 +1,55 @@
...
+            ./ci/travis-failfast.sh
+      - run:
+          name: Unit tests
+          command: |
+            ./ci/travis-unit.sh
+      - run:
+          name: Detect coredumps and print backtraces
+          command: |
+            ./tools/oio-gdb.py
+          when: on_fail
```
murlock: I'll prefer `always` to catch coredump even on success

PR #1761, opened 2019-05-03
Integrity crawler improvements

fvennetier:
##### SUMMARY
Bring various improvements to `oio-crawler-integrity`, required to make it a daemon later.
- Allow it to check all accounts.
- Fix concurrency issues.
- Load `LbClient` from `ConscienceClient` only when necessary.
- Refactor `Tool` class a little.

##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT NAME
- Python API
- CLI
- oio-crawler-integrity

##### SDS VERSION
```
openio 5.0.0.0a2.dev24
```

### DIFF
--- a/oio/crawler/integrity.py
+++ b/oio/crawler/integrity.py
@@ -24,7 +24,6 @@ from oio.common.green import Event, GreenPool, Queue, sleep
 import os
 import csv
 import sys
-import cStringIO
 import argparse
 
 from oio.common import exceptions as exc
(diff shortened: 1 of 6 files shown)


oio/crawler/integrity.py
```diff
@@ -791,9 +825,11 @@ def main():
         request_attempts=args.attempts,
         verbose=True,
     )
-    entries = csv.reader(source, delimiter=' ')
-    for entry in entries:
-        checker.check(Target(*entry))
+    if entries:
+        for entry in entries:
+            checker.check(Target(*entry), recurse=DEFAULT_DEPTH)
```
murlock: Since the Target constructor has been changed (oio/crawler/integrity.py L45), we should add more tests using CSV input

fvennetier: The `Target` constructor argument order has been restored to what it was initially. I don't know if the standard input feature is actually used, but yes, we could add some tests.

PR #1766, opened 2019-05-07
Rdir: Free error on missing options

kamel-rahim:
##### SUMMARY

Since they are not mandatory we should just free the error.

##### ISSUE TYPE

 - Bugfix Pull Request

##### COMPONENT NAME

 - rdir
##### SDS VERSION

```
openio 5.0.0.0a2.dev18
```

### DIFF
--- a/rdir/rdir.c
+++ b/rdir/rdir.c
@@ -2620,6 +2620,8 @@ grid_main_configure(int argc, char **argv)
 	}
 
 	service_id = CFG("service_id");
+	if (err)
+		g_clear_error(&err);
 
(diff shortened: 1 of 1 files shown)


rdir/rdir.c
```diff
@@ -2620,6 +2620,8 @@ grid_main_configure(int argc, char **argv)
 	}
 
 	service_id = CFG("service_id");
+	if (err)
+		g_clear_error(&err);
```
fvennetier: Notice that `g_clear_error` is null safe and the `if (err)` check is redundant.

PR #1768, opened 2019-05-09
crawler: provide informations from rdir to integrity crawler

murlock:
##### SUMMARY
When `openio-admin rawx check` finds a chunk indexed in rdir but missing from the rawx service, the process stops.

This commit enriches the output with information retrieved from rdir.

##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT NAME
OIO

##### SDS VERSION
```
openio 5.0.0.0b0
```


murlock:
Please note that `openio-admin chunk check` is not fixed with this PR:
we should add a step to search on rdir the `cid` and `content`id`

PR #1772, opened 2019-05-14
CLI: Repair corrupted chunk

kamel-rahim:
##### SUMMARY

Repair corrupted chunk

##### ISSUE TYPE

 - Feature Pull Request

##### COMPONENT NAME

 - CLI
##### SDS VERSION

```
openio 5.0.0.0b1.dev3

```

### DIFF
--- a/oio/content/repairer.py
+++ b/oio/content/repairer.py
@@ -19,7 +19,7 @@ from datetime import datetime
 from oio.api.object_storage import _sort_chunks
 from oio.blob.operator import ChunkOperator
 from oio.blob.client import BlobClient
-from oio.common.exceptions import NotFound
+from oio.common.exceptions import NotFound, ClientPreconditionFailed
 from oio.common.storage_method import STORAGE_METHODS
 from oio.common.tool import Tool, ToolWorker
 from oio.common.utils import cid_from_name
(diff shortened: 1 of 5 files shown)


rawx-apache2/src/rawx_repo_core.c
```diff
@@ -479,7 +479,10 @@ chunk_verify_checksum(dav_resource *resource, request_rec *r)
 		EXTRA_ASSERT(resource->info);
 		struct chunk_textinfo_s *p = &resource->info->chunk;
 
-		EXTRA_ASSERT(p->chunk_hash);
+		if (! p->chunk_hash) {
+			g_checksum_free(md5);
+			return APR_EMISMATCH;
```
AymericDu: `user.grid.chunk.hash` is a mandatory xattr. So if the hash isn't available, it's not a `Precondition Failed`. We should return an `Internal Server Error` (for example).

oio/content/repairer.py
```diff
@@ -157,11 +157,13 @@ def _repair_metachunk(self, item, content_id, stg_met, pos, chunks):
             try:
                 self.blob_client.chunk_head(
                     chunk['url'], xattr=True, check_hash=True)
-            except NotFound:
+            except (NotFound, ClientPreconditionFailed) as e:
                 exc = self._safe_chunk_rebuild(
```
AymericDu: `ChunkOperator.rebuild` allows to delete chunk. We could use the `kwargs` to delete chunk safely.
```suggestion
                if isinstance(e, ClientPreconditionFailed):
                    try_chunk_delete = True
                else:
                    try_chunk_delete = False
                exc = self._safe_chunk_rebuild(
                    item, content_id, chunk['url'], try_chunk_delete=try_chunk_delete)
```

AymericDu: A beginner's mistake was made. I updated my suggestion.

PR #1774, opened 2019-05-14
Test 'openio-admin <item> check'

AymericDu:
##### SUMMARY

- Test 'openio-admin account check'
- Test 'openio-admin container check'
- Test 'openio-admin object check'
- Test 'openio-admin chunk check'
- Allow to recover object meta when checking a missing chunk

##### ISSUE TYPE

 - Bugfix Pull Request

##### COMPONENT NAME

- Admin CLI

##### SDS VERSION

```
openio 5.0.0.0b1.dev13
```

### DIFF
--- a/oio/crawler/integrity.py
+++ b/oio/crawler/integrity.py
@@ -32,6 +32,7 @@ from oio.common.logger import get_logger
 from oio.common.storage_method import STORAGE_METHODS
 from oio.api.object_storage import ObjectStorageApi
 from oio.api.object_storage import _sort_chunks
+from oio.rdir.client import RdirClient
 
 DEFAULT_DEPTH = 4
 
@@ -152,6 +153,8 @@ class Checker(object):
             logger=self.logger,
             max_retries=request_attempts - 1,
             request_attempts=request_attempts)
+        self.rdir_client = RdirClient(
+            {"namespace": namespace}, logger=self.logger)
 
         self.accounts_checked = 0
         self.containers_checked = 0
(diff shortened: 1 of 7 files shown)


oio/crawler/integrity.py
```diff
@@ -210,6 +213,28 @@ def complete_target_from_chunk_metadata(self, target, xattr_meta):
                                      "and container names: %s",
                                      cid, err)
 
+    def recover_and_complete_object_meta(self, target, chunk):
+        _, rawx_service, chunk_id = chunk.rsplit('/', 2)
+        # 1. Fetch chunk list from rdir (could be cached).
+        # Unfortunately we cannot seek for a chunk ID.
```
murlock: Is it possible to use `start_after` with value `chunk_id - 1`?

tests/functional/cli/admin/test_item_check.py
```diff
@@ -0,0 +1,1357 @@
...
+        opts = self.get_opts(['Id'])
+        output = self.openio(
+            '--oio-account %s object locate %s --auto --flat-bits %d %s'
+            % (account, obj_name, self.FLAT_BITS, opts))
+        obj_chunks = output.rstrip('\n').split('\n')
+        container = self.autocontainer(obj_name)
+        self._wait_events(account, container, obj_name)
+        return container, obj_meta, obj_chunks
+
+    def corrupt_chunk(self, chunk):
```
murlock: Should we add also `corrupt_chunk_xattr` ?
(Maybe it is done in another PR ?)

oio/crawler/integrity.py
AymericDu: We can't use `start_after`... This parameter uses this format: `CONTAINER_ID|CONTENT_ID|CHUNK_ID`. And we don't know `CONTAINER_ID` and `CONTENT_ID` (because that's what we're looking for).

tests/functional/cli/admin/test_item_check.py
AymericDu: I agree we are doing this test in another PR.

oio/crawler/integrity.py
fvennetier: In the future we can implement a route to make this filtering server-side.

PR #1776, opened 2019-05-16
Improve the service resolver

AymericDu:
##### SUMMARY

Share and improve the class to resolve the service IDs.

##### ISSUE TYPE

 - Bugfix Pull Request

##### COMPONENT NAME

- Python API

##### SDS VERSION

```
openio 4.4.3.dev11
```


oio/common/service_resolver.py
```diff
@@ -0,0 +1,72 @@
...
+from oio.conscience.client import ConscienceClient
+
+
+class ServiceResolver(object):
+    """A caching client to rawx services."""
+
+    def __init__(self, conf, max_age=60, logger=None):
+        self.conf = conf
+        self.logger = logger or get_logger(conf)
+        self.conscience_client = ConscienceClient(self.conf, logger=logger)
```
fvennetier: I have a concern with this class having its own instance of `ConscienceClient` (with its own set of connections to the proxy). Actually I think `ServiceResolver` could be merged with `ConscienceClient`.

oio/common/service_resolver.py
```diff
@@ -0,0 +1,72 @@
...
+            return cached_service_id['addr']
+        result = self.conscience_client.resolve(
+            srv_type=service_type, service_id=service_id)
+        service_addr = result['addr']
+        self._cache[service_id] = {'addr': service_addr,
+                                   'mtime': time.time()}
+        return service_addr
+
+    def resolve_url(self, service_type, url):
+        """
```
kamel-rahim: `resolve_url` is misleading because you also resolve service ID. Just `resolve` should be fine.

oio/common/service_resolver.py
```diff
@@ -0,0 +1,72 @@
...
+        :returns: Resolved URL of a service using a service ID
+        """
+        # FIXME(mb): some tests don't put scheme, should fix tests
+        if not url.startswith('http://'):
+            url = "http://" + url
+
+        parsed = urlparse(url)
+        if parsed.port is not None:
+            return url
```
kamel-rahim: You've removed the `parsed.port == 80` it was your intention?

AymericDu: Yes, it's intentional. There is no reason to make a difference for port 80.

oio/common/service_resolver.py
AymericDu: I am not sure I understand this comment.
- `resolve_url` resolve URLs like `http://OPENIO-rawx-1/66758800F3040241AF884BF82096C0A40FB0E6AFD806C07A2113BBD6442A54D1`
- `resolve_service_id` resolve service IDs like `OPENIO-rawx-1`
I think the comments below these methods help to clarify their use.

kamel-rahim: Yes my bad.

PR #1778, opened 2019-05-16
Do not read conscience persistence file if a hub is running

fvennetier:
##### SUMMARY
If a hub is running, reinjecting the old conscience state can have nasty effects, like respawning service entries that have been explicitly flushed when the local conscience service was down.

Also remove an sqliterepo parameter that is no more used.

##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT NAME
- Consicence
- sqliterepo

##### SDS VERSION
```
openio 4.4.3.dev14
```


fvennetier:
This needs a little more work to either:
- adapt the failing test to retry a few seconds later;
- make the conscience refuse to serve requests until it has fetched a list of services from the other conscience instances.

PR #1779, opened 2019-05-16
Implement 'openio-admin chunk [distributed-]rebuild'

AymericDu:
##### SUMMARY

- Use mixin for all dispatchers
- Implement 'openio-admin chunk [ditributed-]rebuild'

##### ISSUE TYPE

 - Feature Pull Request

##### COMPONENT NAME

- Admin CLI

##### SDS VERSION

```
openio 5.0.0.0b1.dev12
```


oio/cli/admin/item_rebuild.py
```diff
@@ -0,0 +1,128 @@
...
+    Rebuild chunks that were on the specified file.
+    """
+
+
+class ChunkDistributedRebuild(ChunkRebuildCommand):
+    """
+    Rebuild chunks that were on the specified file across the platform.
+    """
+
+    distributed = True
```
kamel-rahim: I don't like this way to split distributed and local in 2 classes.
It would be better to use a proper way (truly object oriented).
But this can be considered as quibbling.

oio/cli/admin/common.py
```diff
@@ -251,3 +251,47 @@ def patch_parser(self, parser):
 
     def check_and_load_parsed_args(self, app, parsed_args):
         pass
+
+
+class ToolCommandMixin(CommandMixin):
+    """
+    Add tool-related argmuments to a cliff command.
```
kamel-rahim: arguments

oio/cli/admin/item_rebuild.py
AymericDu: As the docstrings are different, I have not found any other way.

PR #1781, opened 2019-05-17
CLI: test 'meta2 list containers' and 'rawx list containers'

kamel-rahim:
##### SUMMARY

Test meta2|rawx list containers

##### ISSUE TYPE

 - Tests Pull Request

##### COMPONENT NAME

CLI
##### SDS VERSION

```
openio 5.0.0.0b1.dev11
```

### DIFF
--- a/tests/functional/cli/admin/test_service_list.py
+++ b/tests/functional/cli/admin/test_service_list.py
@@ -1,0 +1,64 @@
+# Copyright (C) 2019 OpenIO SAS, as part of OpenIO SDS
+#
+# This library is free software; you can redistribute it and/or
+# modify it under the terms of the GNU Lesser General Public
+# License as published by the Free Software Foundation; either
+# version 3.0 of the License, or (at your option) any later version.
+#
+# This library is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+# Lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public
+# License along with this library.
+
+from tests.utils import random_str
+from tests.functional.cli import CliTestCase
+
+
+class ServiceListTest(CliTestCase):
+
+    def test_meta2_list_containers(self):
+        container = random_str(16)
+        self.storage.container_create(self.account, container)
+        output = self.storage.directory.list(self.account, container)
+        meta2s = []
+        for srv in output['srv']:
+            if srv['type'] == 'meta2':
+                meta2s.append(srv['host'])
+
+        opts = self.get_format_opts()
+        fullname = self.account + '/' + container
+        for meta2 in meta2s:
+            output = self.openio_admin(
+                'meta2 list containers  %s %s --oio-account %s'
+                % (meta2, opts, self.account))
+            self.assertIn(fullname, output.split('\n'))
+
+        output = self.storage.container_delete(self.account, container)
+        for meta2 in meta2s:
+            output = self.openio_admin(
+                'meta2 list containers %s %s --oio-account %s'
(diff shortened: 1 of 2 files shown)


tests/functional/cli/admin/test_service_list.py
```diff
@@ -0,0 +1,77 @@
...
+                'object create %s %s --name %s %s --oio-account %s'
+                % (container, f.name, obj, opts, self.account))
+
+        self.assertOutput('Ok', output.strip())
+        opts = self.get_format_opts(fields=['Id'])
+        output = self.openio('object locate %s %s %s --oio-account %s'
+                             % (container, obj, opts, self.account))
+        opts = self.get_format_opts(fields=['Name'])
+        fullname = '/'.join([self.account, container])
+        rawx_list = [x[7:-65] for x in output.strip().split('\n')]
```
murlock: I think that a split on first `/` is safe

tests/functional/cli/admin/test_service_list.py
```diff
@@ -0,0 +1,81 @@
...
+from tests.functional.cli import CliTestCase
+
+
+class ServiceListTest(CliTestCase):
+
+    def test_meta2_list_containers(self):
+        opts = self.get_format_opts()
+        container = random_str(16)
+        output = self.openio('container create %s %s --oio-account %s'
+                             % (container, opts, self.account))
```
AymericDu: Since we don't want to test `openio container {create,locate,delete}` and to avoid slowing the test, we should use `ObjectStorageApi.{container_create,container_get_properties,container_delete}`.

tests/functional/cli/admin/test_service_list.py
```diff
@@ -0,0 +1,81 @@
...
+        self.assertOutput('%s True' % container, output.strip())
+        opts = self.get_format_opts(fields=['meta2'])
+        output = self.openio('container locate %s %s  --oio-account %s'
+                             % (container, opts, self.account))
+        meta2s = output.strip().split(' ')
+        opts = self.get_format_opts()
+        fullname = '/'.join([self.ns, self.account, container])
+        for meta2 in meta2s:
+            if meta2.endswith(','):
+                meta2 = meta2[:-1]
```
AymericDu: We should split `output.strip()` with `, ` instead of ` `.

tests/functional/cli/admin/test_service_list.py
```diff
@@ -0,0 +1,81 @@
...
+    def test_rawx_list_containers(self):
+        opts = self.get_format_opts(fields=['Status'])
+        container = random_str(16)
+        obj = random_str(16)
+        with tempfile.NamedTemporaryFile() as f:
+            f.write('test')
+            f.flush()
+            output = self.openio(
+                'object create %s %s --name %s %s --oio-account %s'
+                % (container, f.name, obj, opts, self.account))
```
AymericDu: Since we don't want to test `openio object {create,locate,delete}` and to avoid slowing the test, we should use `ObjectStorageApi.{object_create,object_locate,object_delete}`.

tests/functional/cli/admin/test_service_list.py
```diff
@@ -0,0 +1,81 @@
...
+
+        self.assertOutput('Ok', output.strip())
+        opts = self.get_format_opts(fields=['Id'])
+        output = self.openio('object locate %s %s %s --oio-account %s'
+                             % (container, obj, opts, self.account))
+        opts = self.get_format_opts(fields=['Name'])
+        fullname = '/'.join([self.account, container])
```
AymericDu: If it's possible, I think we should have the same output format as `openio-admin meta2 list containers`, that is `NS/ACCOUNT/CONTAINER`.

fvennetier: Yes, the two commands should output the same format. Whether they should display the namespace name or not, I'm not sure.

PR #1788, opened 2019-05-23
[WIP] oio: Allow to defer the EC computation to a thread pool

jfsmig:
##### SUMMARY
Avoid mixing the I/O intensive workloads of the greenlet-based main thread of python, with the CPU intensive tasks of the EC computation. A GLib-2.0 GThreadPool is used and (linux specific) eventfd synchronization used between them.

This is useful when more than one upload operation happens in the main python thread, that is why the feature has to be explicitly turned on (set `oio.api.storage_method.EC_SYNC` to `False`).

**TODO**: removed the explicit coroutine switching that will happen in #1787 that becomes superfluous (if not counter-productive).
**TODO**: manage errors with adequate exceptions

##### ISSUE TYPE
`enhancement`

##### COMPONENT NAME
`oio`, `ecp`

##### SDS VERSION
`openio 4.4.4.dev14`


jfsmig:
To be rebased on 5.x

PR #1808, opened 2019-06-11
Don't raise an exception when chunk rebuild has worked

AymericDu:
##### SUMMARY

Don't raise an exception when chunk rebuild has worked

##### ISSUE TYPE

 - Bugfix Pull Request

##### COMPONENT NAME

- Python API
- Admin CLI

##### SDS VERSION

```
openio 5.0.0.dev1
```

##### ADDITIONAL INFORMATION

Jira: OS-386

### DIFF
--- a/oio/blob/operator.py
+++ b/oio/blob/operator.py
@@ -14,7 +14,7 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 
-from oio.common.exceptions import ContentNotFound, NotFound, OrphanChunk
+from oio.common.exceptions import ContentNotFound, OrphanChunk
 from oio.common.logger import get_logger
 from oio.content.factory import ContentFactory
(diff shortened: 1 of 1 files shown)


oio/blob/operator.py
```diff
@@ -70,13 +70,19 @@ def rebuild(self, container_id, content_id, chunk_id_or_pos,
...
         # This call does not raise exception if chunk is not referenced
         if chunk_id is not None:
-            self.rdir_client.chunk_delete(
-                chunk.host, container_id, content_id, chunk_id)
+            try:
+                self.rdir_client.chunk_delete(
+                    chunk.host, container_id, content_id, chunk_id)
+            except Exception as exc:
+                self.logger.warn(
+                    'Failed to delete chunk entry (%s) in the rdir (%s): %s',
```
fvennetier: Failed to delete **from** the rdir.

PR #1813, opened 2019-06-14
Update max_versions of container from header (part 2)

murlock:
##### SUMMARY

This commit bring same mecanism to update sys.m2.policy.versions
from header to content/delete operation: we want to control
if current version of object must be deleted or add a delete marker.

Fixes #OS-373

##### ISSUE TYPE
 - Feature Pull Request

##### COMPONENT NAME
proxy and meta2

##### SDS VERSION
```
openio 4.5.2dev4
`

### DIFF
--- a/proxy/m2_actions.c
+++ b/proxy/m2_actions.c
@@ -2934,8 +2934,11 @@ enum http_rc_e action_content_delete (struct req_args_s *args) {
 
 static enum http_rc_e
 _m2_content_delete_many (struct req_args_s *args, struct json_object * jbody) {
+	const char* force_versioning = g_tree_lookup(args->rq->tree_headers,
+			PROXYD_HEADER_FORCE_VERSIONING);
+
 	json_object *jarray = NULL;
-	PACKER_VOID(_pack) { return m2v2_remote_pack_DEL (args->url, NULL, DL()); }
+	PACKER_VOID(_pack) { return m2v2_remote_pack_DEL (args->url, force_versioning, DL()); }
 
 	if (!oio_url_has_fq_container(args->url))
 		return _reply_format_error(args,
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

File: proxy/m2_actions.c [hunk 1]
```diff
@@ -2922,15 +2924,18 @@ enum http_rc_e action_content_show (struct req_args_s *args) {
...
+	PACKER_VOID(_pack) { return m2v2_remote_pack_DEL (args->url, force_versioning, DL()); }
 	GError *err = _resolve_meta2(args, _prefer_master(), _pack, NULL, NULL);
 	return _reply_m2_error (args, err);
 }
 
 static enum http_rc_e
 _m2_content_delete_many (struct req_args_s *args, struct json_object * jbody) {
 	json_object *jarray = NULL;
-	PACKER_VOID(_pack) { return m2v2_remote_pack_DEL (args->url, DL()); }
+	PACKER_VOID(_pack) { return m2v2_remote_pack_DEL (args->url, NULL, DL()); }
```
AymericDu: We should also get the `PROXYD_HEADER_FORCE_VERSIONING` header.

tests/functional/container/test_container.py
```diff
@@ -695,6 +696,12 @@ def test_object_with_versioning_header(self):
         data = json.loads(resp.data)
         self.assertEqual("-1", data['system'].get("sys.m2.policy.version", 0))
 
+        self._delete_content(path, headers={FORCEVERSIONING_HEADER: 1})
+        resp = self.request('POST', self.url_container('get_properties'),
+                            params=params)
+        data = json.loads(resp.data)
+        self.assertEqual("1", data['system'].get("sys.m2.policy.version", 0))
```
AymericDu: We should check that the object has been deleted.

AymericDu: We could add a test where we delete an object with versioning activation.

PR #1814, opened 2019-06-17
Python API: rework EC upload code

fvennetier:
##### SUMMARY
The previous code may have considered as successes some uploads that had failed (removing elements from the list we are iterating on is a bad thing). Now we build a list with only the chunks which do not see any error yet.

Also, parallelize the upload of the trailers, so a slow chunk upload does not block the termination of the other chunk uploads.

##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT NAME
- Python API

##### SDS VERSION
```
openio 4.5.2.dev4
```

### DIFF
--- a/oio/api/ec.py
+++ b/oio/api/ec.py
@@ -743,7 +743,7 @@ class EcMetachunkWriter(io.MetachunkWriter):
         # init generator
         ec_stream.send(None)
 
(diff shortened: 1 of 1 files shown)


racciari:
LGTM

PR #1817, opened 2019-06-21
Send event when changing container properties

AymericDu:
##### SUMMARY

Create a new event `storage.container.update`. This event is sent when changing container properties.

##### ISSUE TYPE

 - Feature Pull Request

##### COMPONENT NAME

- `sqliterepo`
- `meta2`
- `proxy`
- `event-agent`

##### SDS VERSION

```
openio 4.5.3.dev10
```

### DIFF
--- a/sqliterepo/internals.h
+++ b/sqliterepo/internals.h
@@ -89,6 +89,9 @@ struct sqlx_repository_s
 	sqlx_repo_change_hook change_callback;
 	gpointer change_callback_data;
 
+	sqlx_repo_db_properties_change_hook db_properties_change_callback;
+	gpointer db_properties_change_callback_data;
+
 	/* hash for the directory structure */
 	guint hash_width;
 	guint hash_depth;
--- a/sqliterepo/replication_dispatcher.c
+++ b/sqliterepo/replication_dispatcher.c
@@ -2071,10 +2071,12 @@ _handler_PROPSET(struct gridd_reply_ctx_s *reply,
 	for (gchar **p=pairs; !err && *p && *(p+1); p+=2) {
 		if (!(flags & FLAG_NOCHECK)
 				&& !g_str_has_prefix (*p, SQLX_ADMIN_PREFIX_SYS)
-				&& !g_str_has_prefix (*p, SQLX_ADMIN_PREFIX_USER))
+				&& !g_str_has_prefix (*p, SQLX_ADMIN_PREFIX_USER)) {
 			err = NEWERROR(CODE_BAD_REQUEST, "Invalid property name");
+			break;
+		}
 	}
-	if (NULL != err)
+	if (err)
 		goto label_exit;
 
 	/* Open */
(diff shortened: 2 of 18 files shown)


sqliterepo/sqlx_remote.c
```diff
@@ -34,12 +34,17 @@ License along with this library.
 struct asn_TYPE_descriptor_s;
 
 static MESSAGE
-make_request(const gchar *rn, const struct sqlx_name_s *name, gint64 deadline)
+make_request(const gchar *rn, struct oio_url_s *url,
```
kamel-rahim: With this modification are you sure that it should be merged to 4.x ?

AymericDu: If the meta2 is up to date, but the proxy is not up to date, we will receive events of this type:
```
{"event":"storage.container.update","when":1561471129742946,"url":{"ns":"OPENIO"},"request_id":"650AF5C42C0EAD2C7EC0BC0776FF3629","data":{"system":{},"properties":{"test":"1"}}}
```
So during the update, we may have incomplete URLs. But that's not a problem:
- either we need more information and we will fail when updating (no change compared to before)
- or we only need the namespace

sqliterepo/internals.h
```diff
@@ -89,6 +89,9 @@ struct sqlx_repository_s
 	sqlx_repo_change_hook change_callback;
 	gpointer change_callback_data;
 
+	sqlx_repo_send_event_hook send_event_callback;
+	gpointer send_event_callback_data;
```
fvennetier: sqliterepo does not known the event system exist. It cannot decide if an event must be sent or not. The only thing it can do is inform the upper-level that something has changed (in the admin table).

sqliterepo/replication_dispatcher.c
```diff
@@ -1925,34 +1933,55 @@ _handler_PROPDEL(struct gridd_reply_ctx_s *reply,
 	const enum sqlx_open_type_e how = (flags&FLAG_LOCAL)
 		? (SQLX_OPEN_LOCAL|SQLX_OPEN_NOREFCHECK) : SQLX_OPEN_MASTERONLY;
 	err = sqlx_repository_open_and_lock(repo, &n0, how, &sq3, NULL);
+	if (err)
+		goto label_exit;
+
+	if (g_strcmp0(name.type, "meta2") == 0)
+		container_properties = container_properties_new();
```
fvennetier: We cannot implement meta2 code in the sqliterepo library!

sqliterepo/sqlite_utils.c
```diff
@@ -500,3 +502,69 @@ version_extract_from_admin_tree(GTree *t)
 	return v;
 }
 
+struct container_properties_s
+{
+	GTree *system;
+	GTree *user;
+};
```
fvennetier: "Container" is a meta2 thing. Container-related code should not be implemented in sqliterepo.

PR #1818, opened 2019-06-24
Python API: propagate logger instance

murlock:
##### SUMMARY

This PR allows to use logger from `object_create` / `object_fetch` in low level API

##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT NAME
oio

##### SDS VERSION
```
openio 4.5.3
```

### DIFF
--- a/oio/api/ec.py
+++ b/oio/api/ec.py
@@ -34,7 +34,7 @@ from oio.common.constants import CHUNK_HEADERS
 from oio.common.logger import get_logger
 
 
-logger = get_logger({}, __name__)
+LOGGER = get_logger({}, __name__)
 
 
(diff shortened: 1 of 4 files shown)


oio/api/ec.py
```diff
@@ -206,7 +208,7 @@ class ECStream(object):
     Handles the different readers.
     """
     def __init__(self, storage_method, readers, range_infos, meta_length,
-                 fragment_length, reqid=None, perfdata=None):
+                 fragment_length, reqid=None, perfdata=None, logger_=None):
```
fvennetier: This is confusing. I suggest renaming the global logger with upper-case characters, and keep `logger` for the keyword argument.

PR #1825, opened 2019-07-03
Allow lifecycle to work with sharded buckets

murlock:
##### SUMMARY

This PR will allow LifeCycle to be applied on bucket with Container Hierarcht=y

This behavior is triggered by flag on command line.
If set, a container listing will be triggered with `container%2F` as prefix.

The object listing will enrich metadata of objects to contains
original name in current container, computed name (with path)
and current container.

All tests against filter are unchanged, but API calls will use
`org_name` and `container's object`

##### ISSUE TYPE
 - Feature Pull Request

##### COMPONENT NAME
Python LifeCycle

##### SDS VERSION
```
4.6.1dev3
```

### DIFF
--- a/oio/container/lifecycle.py
+++ b/oio/container/lifecycle.py
@@ -1,4 +1,4 @@
-# Copyright (C) 2017 OpenIO SAS, as part of OpenIO SDS
+# Copyright (C) 2017-2019 OpenIO SAS, as part of OpenIO SDS
 #
 # This library is free software; you can redistribute it and/or
 # modify it under the terms of the GNU Lesser General Public
@@ -26,6 +26,7 @@ from oio.common.exceptions import OioException
 from oio.common.logger import get_logger
 from oio.common.utils import depaginate
 from oio.common.easy_value import true_value
+from oio.common.constants import CH_ENCODED_SEPARATOR, CH_SEPARATOR
 
 
 ALLOWED_STATUSES = ['enabled', 'disabled']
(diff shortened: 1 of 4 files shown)


oio/container/lifecycle.py
```diff
@@ -182,28 +184,27 @@ def apply(self, obj_meta, **kwargs):
...
-                container=self.container,
+                container=container,
                 properties=True,
                 versions=True,
                 **kwargs):
             try:
+                # save name inside container for API use
+                obj_meta['org_name'] = obj_meta['name']
+                # and reconstruct full path of object if needed
+                if '%2F' in container:
```
fvennetier: The delimiter should be a variable (or at least a constant). For specific use cases we may want to change it.

AymericDu: We should also check that `container_hierarchy` is enabled.

oio/container/lifecycle.py
```diff
@@ -777,6 +807,7 @@ def __init__(self, filter_, policy, **kwargs):
         super(Transition, self).__init__(filter_, **kwargs)
         self.policy = policy
         if self.lifecycle:
+            print("##### INVESTIGATE HERE")
```
fvennetier: This looks like unfinished work...

oio/container/lifecycle.py
murlock: :+1:

PR #1832, opened 2019-07-11
Integrity crawler: fix rebuilder input

kamel-rahim:
##### SUMMARY

Fix error when `--output-for-blob-rebuilder` is used with only object checking.

##### ISSUE TYPE

 - Bugfix Pull Request

##### COMPONENT NAME

oio-crawler-integrity
##### SDS VERSION

```
openio 4.6.0
```

##### ADDITIONAL INFORMATION

```
Jira: OS-423
```

### DIFF
--- a/oio/crawler/integrity.py
+++ b/oio/crawler/integrity.py
@@ -36,6 +36,9 @@ from oio.blob.client import BlobClient
 from oio.api.object_storage import _sort_chunks
 
 
+IRREPARABLE_PREFIX = '#IRREPARABLE'
+
+
 class Target(object):
     def __init__(self, account, container=None, obj=None, chunk=None):
         self.account = account
@@ -126,7 +129,7 @@ class Checker(object):
     def write_error(self, target, irreparable=False):
         error = list()
         if irreparable:
-            error.append('#IRREPARABLE')
+            error.append(IRREPARABLE_PREFIX)
         error.append(target.account)
         if target.container:
             error.append(target.container)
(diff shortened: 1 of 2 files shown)


oio/crawler/integrity.py
```diff
@@ -122,11 +123,15 @@ def write_error(self, target, irreparable=False):
...
         try:
-            cid = ct_meta['system']['sys.name'].split('.', 1)[0]
+            ct_meta = self.list_cache[(target.account, target.container)][1]
+            try:
+                cid = ct_meta['system']['sys.name'].split('.', 1)[0]
+            except KeyError:
+                cid = ct_meta['properties']['sys.name'].split('.', 1)[0]
         except KeyError:
-            cid = ct_meta['properties']['sys.name'].split('.', 1)[0]
+            cid = cid_from_name(target.account, target.container)
```
fvennetier: I "fixed" this with another solution in #1830. See https://github.com/open-io/oio-sds/pull/1830/files#diff-f925e1e5526264e38481ecb25a4486dbR58

tests/functional/crawler/test_crawler_integrity.py
```diff
@@ -0,0 +1,75 @@
...
+        self.api.container_flush(self.account, self.container)
+        self.api.container_delete(self.account, self.container)
+        self.api.account_delete(self.account)
+
+    def _verify_rebuilder_input(self):
+        line = fileinput.input(self.rebuild_file).next().strip()
+        cid = cid_from_name(self.account, self.container)
+        expected = '|'.join([cid, self.meta['id'], self.chunk['real_url']])
+        self.assertEqual(expected, line)
+        fileinput.close()
```
fvennetier: This line should be in a `finally` block.

PR #1835, opened 2019-07-12
Better integrity crawler

fvennetier:
##### SUMMARY
- Implement new parameters to control when the tool reports problems:
  - after a specified number of confirmations,
  - after a specified delay.
- Allow the tool to wait between passes.
- Send broken object events to a user-selected queue, for example the input queue of `oio-blob-rebuilder`. It is advised to also specify a delay before sending events, otherwise temporary failures may trigger unwanted chunk rebuilds.

##### ISSUE TYPE
 - Feature Pull Request

##### COMPONENT NAME
- `oio-crawler-integrity`

##### SDS VERSION
```
openio 4.6.1.dev15
```

### DIFF
--- a/oio/blob/converter.py
+++ b/oio/blob/converter.py
@@ -20,11 +20,10 @@ import os
 import tempfile
 from string import hexdigits
 from datetime import datetime
-from collections import OrderedDict
 
 from oio.common.constants import chunk_xattr_keys, OIO_VERSION, \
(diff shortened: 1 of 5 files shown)


fvennetier:
> When we use `delayed_targets` without daemon mode, the broken items are not written to output.
Yes, hence the warning in the documentation.

PR #1840, opened 2019-07-18
Allow to simulate versioning when creating and deleting an object

AymericDu:
##### SUMMARY

Allow to simulate versioning when creating and deleting an object

##### ISSUE TYPE

 - Feature Pull Request

##### COMPONENT NAME

- `proxy`
- `meta2`

##### SDS VERSION

```
openio 4.6.1.dev16
```


murlock:
We should add a note to avoid using `force_versioning` and `simulate_versioning` at same time IMO

AymericDu:
> We should add a note to avoid using `force_versioning` and `simulate_versioning` at same time IMO

We can use `force_versioning` and `simulate_versioning` at same time. It will simulate the versioning and change its actual value.

PR #1841, opened 2019-07-18
Allow to add delete marker to a specific version

AymericDu:
##### SUMMARY

Allow to add delete marker to a specific version

##### ISSUE TYPE

 - Feature Pull Request
 - Bugfix Pull Request

##### COMPONENT NAME

- `proxy`
- `meta2`

##### SDS VERSION

```
openio 4.6.1.dev19
```


tests/functional/container/test_container.py
```diff
@@ -787,6 +790,76 @@ def test_object_with_versioning_header_with_delete_many(self):
...
+        self._create_content(path)
+        props = {"system": {"sys.m2.policy.version": '-1'}}
+        resp = self.request('POST', self.url_container('set_properties'),
+                            params=params_container,
+                            data=json.dumps(props))
+        self.assertEqual(200, resp.status)
+
+        self._create_content(path, version=12345)
+        self._create_content(path, version=12345, create_status=409)
+
```
murlock: Does it make sense to  test creation of a delete marker of object at version 12345 ?

AymericDu: I already add a test for this case: https://github.com/open-io/oio-sds/pull/1841/commits/0b940c22ef5509f3eb7b8f6de2c1dde7fbc42665#diff-b840da077c2e6f9754bf0e5c70ebb72cR820

meta2v2/meta2_utils.c
```diff
@@ -1181,28 +1232,65 @@ m2db_delete_alias(struct sqlx_sqlite3_s *sq3, gint64 max_versions,
...
 
+	gboolean add_delete_marker = FALSE;
 	if (VERSIONS_DISABLED(max_versions) || VERSIONS_SUSPENDED(max_versions) ||
 			oio_url_has(url, OIOURL_VERSION) || ALIASES_get_deleted(alias)) {
-		err = _real_delete_and_save_deleted_beans(sq3, beans, alias, header,
-				cb, u0);
+		if (delete_marker) {
+			if (VERSIONS_DISABLED(max_versions) ||
+					VERSIONS_SUSPENDED(max_versions)) {
+				err = NEWERROR(CODE_BAD_REQUEST,
```
jfsmig: The `BADREQ` macro might help you to keep the lines short enough to avoid folding.

AymericDu:
> On purpose itself of the current PR, are we allowed to choose whether or not a Deletion marker is opportune? Isn't it mandatory by the (s3) norm?

Indeed, the norm does not allow to create a delete marker on a specific version. We can only create one on the latest version.
But `oio-replicator` has no guarantee on the order of events, we must have a way to create a delete marker on a specific version.

PR #1845, opened 2019-07-25
mover: Allow to move only chunk whose size is in a range

jfsmig:
##### SUMMARY
Allow to move chunks  only if their size is within a configured range.
Also wait for pending movers to complete before exiting a moving round.

##### ISSUE TYPE
enhancement

##### COMPONENT NAME
oio-blob-mover

##### SDS VERSION
`openio 4.6.1.dev24`

### DIFF
--- a/oio/blob/mover.py
+++ b/oio/blob/mover.py
@@ -149,6 +149,7 @@ class BlobMoverWorker(object):
             mover_time += (now - loop_time)
             if self.limit != 0 and self.total_chunks_processed >= self.limit:
                 break
+        pool.waitall()
         elapsed = (time.time() - start_time) or 0.000001
         self.logger.info(
             '%(elapsed).02f '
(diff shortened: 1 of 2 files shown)


oio/blob/mover.py
```diff
@@ -193,6 +194,18 @@ def chunk_move(self, path, chunk_id):
         content_id = meta['content_id']
         chunk_id = meta['chunk_id']
 
+        # Maybe skip the chunk because it doesn't match the size constaint
+        chunk_size = int(meta['chunk_size'])
+        min_chunk_size = int(self.conf.get('min_chunk_size', 0))
+        max_chunk_size = int(self.conf.get('max_chunk_size', 0))
```
AymericDu: We could initialize attributes in the constructor for these values.

jfsmig: This is intended to be a minimal change. If a wrong configuration value is set (e.g. a string), we behave *much* better like this because the exception is cleanly catched and the error is cleanly reported. If the value is parsed in the constructor, we behave much worse (we loop and fill the logs with ugly backtraces).

jfsmig:
> With the `ratelimit` (https://github.com/open-io/oio-sds/pull/1845/files#diff-2e33b6b9274ea3f6539476d37c5a6135L117), we will wait without doing anything when the size of the chunk will not be in between.

If we check the size before rate-limiting, the risk is that an oio-blob-mover that runs on a rawx full of small chunks to crazy-loop on the filesystem and generate a huge load for nothing. the rate-limiting is also here to protect the local volume.

PR #1849, opened 2019-07-31
Set an operation-dependent prefix on request ids

murlock:
##### SUMMARY
Add more prefix for request-id to identify operation

Some debugs logs will be emitted with EXTRA_DEBUG when proxy receives request without reqid or reqid is generated without prefix

##### ISSUE TYPE
 - Feature Pull Request

##### COMPONENT NAME
sds

##### SDS VERSION
```
5.0.1dev4
```

##### ADDITIONAL INFORMATION

We have still investigate why DB_USE from autocreate operation doesn't reuse reqid from original request.

### DIFF
--- a/oio/api/object_storage.py
+++ b/oio/api/object_storage.py
@@ -30,7 +30,8 @@ from oio.common.utils import cid_from_name, GeneratorIO, monotonic_time, \
     depaginate, set_deadline_from_read_timeout
 from oio.common.easy_value import float_value, true_value
 from oio.common.logger import get_logger
-from oio.common.decorators import ensure_headers, ensure_request_id
+from oio.common.decorators import ensure_headers, ensure_request_id, \
+    ensure_request_id2
 from oio.common.storage_method import STORAGE_METHODS
 from oio.common.constants import OIO_VERSION, HEADER_PREFIX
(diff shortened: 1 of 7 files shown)


oio/api/object_storage.py
```diff
@@ -705,6 +706,7 @@ def object_create_ext(self, account, container, file_or_path=None,
         """
         if (data, file_or_path) == (None, None):
             raise exc.MissingData()
+        kwargs['reqid'] = request_id('create-')
```
fvennetier: Unfortunately this overwrites the request ID set by the client application or the method decorator.

murlock: Done

oio/api/object_storage.py
```diff
@@ -705,6 +706,7 @@ def object_create_ext(self, account, container, file_or_path=None,
         """
         if (data, file_or_path) == (None, None):
             raise exc.MissingData()
+        kwargs.setdefault('reqid', request_id('create-'))
```
fvennetier: At this point, there is already a request ID, set by the `ensure_request_id`. If you want the generated request ID to be prefixed by something, replace the decorator by `ensure_request_id2` and pass the prefix as an argument.

PR #1851, opened 2019-08-01
event-agent: Permit to not replicate some accounts and containers

kamel-rahim:
##### SUMMARY

Permit to exclude account and containers from replication events.
For testing purpose and check tools we do not want to replicate some accounts and containers
to the other namespace.

##### ISSUE TYPE

 - Feature Pull Request

##### COMPONENT NAME

event-agent
##### SDS VERSION

```
openio 4.6.1.dev26
```
##### ADDITIONAL INFORMATION
```
Jira: OS-433
```


tests/functional/event/filters/test_filters.py
```diff
@@ -365,3 +365,36 @@ def test_missing_m1_chunk_ec(self):
...
+            [quote_plus('account'), 'account2/container'])
+        self.assertFalse(self.notify_filter._should_replicate('account', random_str(16)))
+        self.assertFalse(self.notify_filter._should_replicate('account2', 'container'))
+
+        # account that should not be replicated
+        strange_account = [quote_plus(x) for x in strange_paths]
+        self.notify_filter.no_replicate = self.notify_filter._parse_no_replicate(strange_account)
+        print self.notify_filter.no_replicate
+        for x in strange_paths:
+            print x
```
murlock: To be removed

tests/functional/event/filters/test_filters.py
```diff
@@ -365,3 +365,36 @@ def test_missing_m1_chunk_ec(self):
...
+        # Simple test
+        self.notify_filter.no_replicate = self.notify_filter._parse_no_replicate(
+            [quote_plus('account'), 'account2/container'])
+        self.assertFalse(self.notify_filter._should_replicate('account', random_str(16)))
+        self.assertFalse(self.notify_filter._should_replicate('account2', 'container'))
+
+        # account that should not be replicated
+        strange_account = [quote_plus(x) for x in strange_paths]
+        self.notify_filter.no_replicate = self.notify_filter._parse_no_replicate(strange_account)
+        print self.notify_filter.no_replicate
```
murlock: To be removed

oio/event/filters/notify.py
```diff
@@ -32,12 +70,16 @@ def init(self):
 
     def process(self, env, cb):
         data = json.dumps(env)
```
murlock: We should move l.72 inside the `if`
It may make some differences if there is a lot of account/container to skip

oio/event/filters/notify.py
```diff
@@ -22,8 +23,45 @@
 
 class NotifyFilter(Filter):
 
+    def _parse_no_replicate(self, array):
```
murlock: Could be a static method ?

oio/event/filters/notify.py
```diff
@@ -22,8 +23,45 @@
...
+        elif container in containers :
+            return False
+        else:
+            return True
+
     def init(self):
         queue_url = self.conf.get('queue_url')
+        if self.conf.get('no_replicate', None):
+            self.no_replicate = self._parse_no_replicate(
+                json.loads(self.conf.get('no_replicate')))
```
murlock: A default value should be added: `[]`

oio/event/filters/notify.py
```diff
@@ -22,22 +23,63 @@
...
+            return True
+        elif container in containers:
+            return False
+        else:
+            return True
+
     def init(self):
         queue_url = self.conf.get('queue_url')
+        self.no_replicate = self._parse_no_replicate(
+            self.conf.get('no_replicate', []))
```
fvennetier: I think "exclude" would be more explicit than "no_replicate".
Also, if `conf` is loaded from a file, there is a chance that the value is a string and not a list.

oio/event/filters/notify.py
```diff
@@ -13,6 +13,7 @@
 # You should have received a copy of the GNU Affero General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
+from urllib import unquote_plus
```
fvennetier: I think we have had problems with `quote_plus`, I suggest using `quote`.

PR #1855, opened 2019-08-09
Prepare chunk locations in the proxy instead of the meta2

jfsmig:
##### SUMMARY
Avoid hitting meta2 when asking for chunks locations.

##### ISSUE TYPE
enhancement

##### COMPONENT NAME
`*`

##### SDS VERSION
`openio 5.0.2.dev45`


tests/functional/container/test_container.py
```diff
@@ -1119,6 +1121,7 @@ def _test_spare_with_n_broken(self, count_broken):
                             data=json.dumps(
                                 {"notin": chunks,
                                  "broken": broken}))
+        print resp.data
```
AymericDu: An oversight, I think.

PR #1864, opened 2019-08-26
Wait for proxy and conscience when waiting for services

AymericDu:
##### SUMMARY

Wait for services even if the proxy and conscience are not ready yet.

##### ISSUE TYPE

 - Bugfix Pull Request

##### COMPONENT NAME

- `proxy`
- CLI

##### SDS VERSION

```
openio 4.7.0.0rc1.dev7
```

### DIFF
--- a/proxy/common.c
+++ b/proxy/common.c
@@ -35,15 +35,22 @@ validate_namespace (const char * ns)
 	return 0 == strcmp (ns, ns_name);
 }
 
-gboolean
-validate_srvtype (const char * n)
+GError *
+validate_srvtype(const char *srvtype)
 {
-	gboolean rc = FALSE;
-	NSINFO_READ(if (srvtypes) {
-		for (gchar ** p = srvtypes; !rc && *p; ++p)
-			rc = !strcmp (*p, n);
-	});
-	return rc;
+	gboolean service_types_loaded = FALSE;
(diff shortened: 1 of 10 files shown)


proxy/common.h
```diff
@@ -134,6 +134,7 @@ extern gchar **srvtypes;
 extern struct namespace_info_s nsinfo;
 gboolean validate_namespace (const char * ns);
 gboolean validate_srvtype (const char * n);
+GError * validate_srvtype_v2 (const char * n);
```
murlock: You should add a small explanation between validate_srvtype and validate_srvtype_v2

proxy/common.c
```diff
@@ -36,16 +36,35 @@ validate_namespace (const char * ns)
...
+	NSINFO_READ(
+		if (srvtypes) {
+			service_types_loaded = TRUE;
+			for (gchar ** p = srvtypes; !service_type_found && *p; ++p)
+				service_type_found = !strcmp(*p, srvtype);
+		});
+	if (!service_types_loaded)
+		return NEWERROR(CODE_UNAVAILABLE, "Service types not yet loaded");
+	if (!service_type_found)
+		return BADSRVTYPE();
```
fvennetier: In the future I think we should modify this macro so it prints the name of the "unexpected" service type.

PR #1866, opened 2019-08-27
rawx: Update the log format

jfsmig:
##### SUMMARY
* Only one format and semantics for any access log entry 
* Display both `bytes_in` and `bytes_out` for an easier integration
* Stop displaying the container id.

##### ISSUE TYPE
enhancement

##### COMPONENT NAME
rawx

##### SDS VERSION
``

### DIFF
--- a/rawx/logger.go
+++ b/rawx/logger.go
@@ -124,34 +124,43 @@ func LogDebug(format string, v ...interface{}) {
 	writeLogFmt(syslog.LOG_DEBUG, format, v...)
 }
 
-func LogHttp(url, peer, method string, status int, spent, bytes uint64, containerID, id, path string) {
+type AccessLogEvent struct {
+	status    int
+	bytesIn   uint64
+	bytesOut  uint64
+	timeSpent uint64
+	method    string
+	local     string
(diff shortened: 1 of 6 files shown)


rawx/logger.go
```diff
@@ -124,34 +124,43 @@ func LogDebug(format string, v ...interface{}) {
...
 	sb.WriteRune(' ')
-	if len(containerID) > 0 {
-		sb.WriteString(containerID)
-	} else {
-		sb.WriteRune('-')
-	}
-	sb.WriteRune(' ')
-	sb.WriteString(id)
+	sb.WriteString(utoa(evt.bytesIn))
+	sb.WriteString(" - ")
```
fvennetier: It looks like `bytesIn` replaces `containerID`, so why do we print this dash?

vdombrovski: On the deployment/provisioning side, we chose to not introduce any breaking changes here, as these will mostly become silent errors anyway. (see referenced PR)

fvennetier: reqId and path have been shifted right, isn't that a breaking change?

PR #1872, opened 2019-08-30
cmake: Only install prod-related tools in other build types than Debug

jfsmig:
##### SUMMARY
The tools related to development tasks is only installed if the suite is built with `CMAKE_BUILD_TYPE=Debug`. It will avoid deploying `oio-reset.sh` (and its friends) with any package.

##### ISSUE TYPE
simplification

##### COMPONENT NAME
cmake

##### SDS VERSION
`openio 5.1.1.dev58`

### DIFF
--- a/tools/CMakeLists.txt
+++ b/tools/CMakeLists.txt
@@ -40,13 +40,21 @@ configure_file(${CMAKE_CURRENT_SOURCE_DIR}/oio-sds.pc.in
 configure_file(${CMAKE_CURRENT_SOURCE_DIR}/oio-bootstrap.py
 	${CMAKE_CURRENT_BINARY_DIR}/oio-bootstrap.py @ONLY)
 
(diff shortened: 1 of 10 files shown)


tools/CMakeLists.txt
```diff
@@ -40,13 +40,21 @@ configure_file(${CMAKE_CURRENT_SOURCE_DIR}/oio-sds.pc.in
...
 
-install(TARGETS oio-tool DESTINATION bin)
+install(TARGETS
+			oio-tool
+		DESTINATION bin)
 
-install(TARGETS oio-file DESTINATION bin)
+install(TARGETS
+			oio-file
+			oio-lb-benchmark
```
fvennetier: This tool is still a bit rude, but it can help admins test their service pool configuration. I vote for it to be installed on production environments.

PR #1873, opened 2019-08-30
Avoid overriding NS config from the system with local files

jfsmig:
##### SUMMARY
If system-wide files have been loaded, avoid using the local files in the C and Python code.

##### ISSUE TYPE
bugfix

##### COMPONENT NAME
core, python SDK

##### SDS VERSION
`openio 5.1.1.dev58`


core/cfg.c
```diff
@@ -38,22 +38,32 @@ oio_cfg_build_key(const gchar *ns, const gchar *what)
...
+ * `override` is only set when loading local files. This happens in testing
+ * conditions and we prefer avoiding any overriding of NS config.
+ */
 static void
-config_load_ns(GHashTable *h, GKeyFile *kf, const gchar *ns)
+config_load_ns(GHashTable *h, GKeyFile *kf, const gchar *ns, gboolean override)
 {
 	gchar **keys = g_key_file_get_keys(kf, ns, 0, NULL);
 	if (keys) {
+		gchar *k_known = oio_cfg_build_key(ns, "known");
```
fvennetier: Not freed when not found in hash table.

PR #1875, opened 2019-09-02
Check for a quorum when linking objects

AymericDu:
##### SUMMARY

Use a quorum to link an object

##### ISSUE TYPE

 - Bugfix Pull Request

##### COMPONENT NAME

- Python API
- `proxy`

##### SDS VERSION

```
openio 4.7.0.0rc1.dev10
```


oio/api/object_storage.py
```diff
@@ -456,7 +456,7 @@ def container_snapshot(self, account, container, dst_account,
                 fullpath = encode_fullpath(
                     dst_account, dst_container, obj['name'], obj['version'],
                     obj['content'])
-                storage_method = STORAGE_METHODS.load(obj['chunk_method'])
+                storage_method = STORAGE_METHODS.load(obj['chunk-method'])
```
murlock: Other parts use `chunk_method` (for example https://github.com/open-io/oio-sds/blob/309db2cedf8093aaa710caf5aca84559c7ad86a7/oio/api/object_storage.py#L1110), it seems confusing to mix both of them

oio/api/io.py
```diff
@@ -79,7 +79,73 @@ def readinto(self, b):  # pylint: disable=invalid-name
...
+    def link(self):
+        content_chunks = list()
+
+        kwargs = MetachunkLinker.filter_kwargs(self.extra_kwargs)
+        for meta_chunk in self.chunk_prep():
+            try:
+                handler = MetachunkLinker(
+                    meta_chunk, self.fullpath, self.blob_client,
+                    storage_method=self.storage_method,
+                    reqid=self.headers.get('X-oio-req-id'),
```
murlock: Please use `REQID_HEADER` const

PR #1878, opened 2019-09-04
Implement oio-admin-agent

vdombrovski:
### SUMMARY

This agent provides an HTTP API for running various long-term admin tasks from each node. It can only run jobs locally for local services (defined at launch via --host, and by parsing location).

##### Quick API Summary:

```
POST /api/v1/jobs - create a new job
DELETE /api/v1/jobs/[job_id] - terminate a running job
GET /api/v1/jobs - list jobs
```

##### Job structure:

```
{"action": "rebuild|move", "type": "rawx|meta2", "src": "IP:PORT", opts**}
```
> Additional runtime options are tied to each job type.
>
> Blob mover supports:
> - bps (bytes per sec)
> - cps (chunks per sec)
> - concurrency
> - target
> - minsize
> - maxsize
> - exclude (list of IP:PORT or `re:{!}{expr}` (see tests for example)
>
> Meta2 mover supports:
> - concurrency (by the agent itself, not the Meta2Mover)
> - minsize (minimal size of bases to move)
> - maxsize (max size of bases to move)
>
> Blob Rebuilder doesn't support any additional options currently

##### Available jobs:

- blob_mover
- blob_rebuilder (distributed)
- meta2_mover

This also provides a temporary binary, oio-mover-client, that can interact with the admin-agent.

### ISSUE TYPE
 - Feature Pull Request

### COMPONENT NAME
- oio-admin-agent

### SDS VERSION
```
openio 5.0.2.dev2
```

### ADDITIONAL INFORMATION

Tests can be launched with:

```
nosetests tests/unit/admin/test_admin_agent.py
nosetests tests/functional/admin/test_admin_agent.py
```


oio/admin/agent.py
```diff
@@ -0,0 +1,570 @@
...
+            return to_exclude
+
+    def chunk_bases(self, bases, into=1):
+        """
+        Chunk meta2 bases into N chunks to allow
+        for mover parallelization
+        """
+        chunks = [bases[i::into] for i in range(into)]
+        for i, c in enumerate(chunks):
+            chunks[i] = [(k, v) for k, v in dict(c).iteritems()]
```
fvennetier: I don't know which type is `bases`, but I suppose this can be simplified like
```suggestion
            chunks[i] = dict(c).items()
```

oio/admin/agent.py
```diff
@@ -0,0 +1,570 @@
...
+
+        def _add(lock_, field, value):
+            lock_.acquire()
+            field.value += value
+            lock_.release()
+
+        lock = control.get('lock')
+        src = config.get('src')
+        conf_str = ",".join(["%s=%s" % (k, v) for k, v in config.items()])
+        self.client.lock_score(dict(type="meta2", addr=src))
```
fvennetier: Declaring dictionaries this way make things harder to read. I suggest this syntax: `{'type': 'meta2', 'addr': 'src'}`.

oio/admin/agent.py
```diff
@@ -0,0 +1,570 @@
...
+
+class BaseAdminAgentHandler(BaseHTTPRequestHandler):
+    """
+    Mover agent handler
+    Handles incoming HTTP requests
+    """
+    agent = None
+
+    def __init__(self, *args, **kwargs):
+        BaseHTTPRequestHandler.__init__(self, *args, **kwargs)
```
fvennetier: No need to declare `__init__` if the only statement is calling the superclass' constructor.
BTW, prefer calling `super(BaseAdminAgentHandler, self).__init__(*args, **kwargs)`.

oio/admin/agent.py
```diff
@@ -0,0 +1,570 @@
...
+            pass
+        self.cm = OioAdminAgentClassManager(ctx=ctx or kwargs.get("ctx", {}))
+        self.host = args.host
+        self.namespace = args.namespace
+        self.client = self.cm.conscience({"namespace": args.namespace})
+        self.log = get_logger_from_args(args)
+        self.log.info("Starting oio-mover-agent")
+        self.jobs = dict()
+
+        signal.signal(signal.SIGTERM, self._clean_exit)
```
vdombrovski: SIGINT

bin/oio-admin-agent
```diff
@@ -0,0 +1,58 @@
...
+def make_arg_parser():
+    desc = """
+        Stateless daemon providing an HTTP interface to run blob/meta2 movers
+    """
+    parser = argparse.ArgumentParser(description=desc,
+                                     parents=[make_logger_args_parser()])
+    parser.add_argument('--namespace',
+                        metavar='<namespace>',
+                        dest='namespace',
+                        help="Namespace", required=True)
```
fvennetier: Could we use short argument names, or better: only positional arguments (all 3 are mandatory).

oio/admin/agent.py
```diff
@@ -0,0 +1,570 @@
...
+                return True
+
+    def volume(self, type_, src):
+        """
+        Resolve the volume for the specified service
+        """
+        for svc in self.client.all_services(type_):
+            tags = svc.get("tags", {})
+            location = tags.get("tag.loc")
+            if svc.get("addr") == src and self.host in location.split("."):
```
fvennetier: I think we need to match service IDs, not addresses.

tests/functional/admin/test_admin_agent.py
```diff
@@ -0,0 +1,152 @@
...
+                blob_rebuilder=FakeBlobRebuilderSlow,
+                meta2_mover=FakeMeta2MoverSlow,
+            )
+        ))
+        httpd = HTTPServer(('localhost', 0), self.handler)
+        self.url = "http://localhost:%s/" % httpd.server_port
+
+        def start(httpd):
+            httpd.serve_forever()
+        global process
```
murlock: Why global since it doesn't use anywhere else only in tearDown ?
Maybe you should register it in `setUpClass` ?

tools/oio-mover-client.go
```diff
@@ -0,0 +1,481 @@
+/*
```
murlock: At least it should be build in `travis`

tools/oio-mover-client.go
```diff
@@ -0,0 +1,481 @@
...
+            }
+        default:
+            // TODO: Use availActions here
+            log.Fatalln("Invalid action " + action + ", must be in (start|status|stop)")
+    }
+}
+
+func parseStatus(status int64) (string) {
+    switch status {
+        case 0:
```
murlock: please use `const`

fvennetier:
Hi. I think this has been superseded by xcute. Can we close the pull-request?

PR #1884, opened 2019-09-06
Improve oio-blob-registrator

AymericDu:
##### SUMMARY

- Insert alias and content in the containers
- Allow to specify containers

##### ISSUE TYPE

 - Bugfix Pull Request

##### COMPONENT NAME

- `oio-blob-registrator`

##### SDS VERSION

```
openio 4.7.1.dev8
```

### DIFF
--- a/meta2v2/meta2_utils_json_in.c
+++ b/meta2v2/meta2_utils_json_in.c
@@ -40,10 +40,10 @@ m2v2_json_load_single_alias (struct json_object *j, gpointer *pbean)
 	struct oio_ext_json_mapping_s m[] = {
 		{"name",    &jname,    json_type_string,  1},
 		{"version", &jversion, json_type_int,     1},
-		{"header",  &jheader,  json_type_string,  1},
 		{"ctime",   &jctime,   json_type_int,     0},
 		{"mtime",   &jmtime,   json_type_int,     0},
 		{"deleted", &jdel,     json_type_boolean, 0},
+		{"header",  &jheader,  json_type_string,  1},
 		{NULL, NULL, 0, 0}
 	};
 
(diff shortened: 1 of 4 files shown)


meta2v2/meta2_utils_json_in.c
```diff
@@ -79,13 +81,15 @@ m2v2_json_load_single_header (struct json_object *j, gpointer *pbean)
...
+	struct json_object *jid, *jhash, *jsize, *jctime, *jmtime, *jpolicy,
+			*jmethod, *jtype;
 	struct oio_ext_json_mapping_s mapping[] = {
-		{"id",     &jid,    json_type_string, 1},
-		{"hash",   &jhash,  json_type_string, 1},
-		{"size",   &jsize,  json_type_int, 1},
-		{"ctime",  &jctime, json_type_int, 0},
-		{"mtime",  &jmtime, json_type_int, 0},
+		{"id",           &jid,     json_type_string, 1},
+		{"hash",         &jhash,   json_type_string, 0},
```
murlock: The hash is no more mandatory, I think you should document why hash is no more mandatory here

oio/blob/registrator.py
```diff
@@ -14,181 +14,273 @@
...
+        self.last_report = 0
+        self.report_interval = int_value(
+            conf.get('report_interval'),
+            self.DEFAULT_REPORT_INTERVAL)
+
+        self.client = ContainerClient(
+            {'namespace': self.namespace}, logger=self.logger)
+        self.ctime = int(time.time())
+
+    def _beans_from_meta(self, meta):
```
murlock: Maybe it could be factorized with `meta2beans` in first commit ?

AymericDu: It's the same method, but I've renamed it.

PR #1891, opened 2019-09-19
Test the download with ECD

AymericDu:
##### SUMMARY

Test the download with ECD

##### ISSUE TYPE

 - Test Pull Request

##### COMPONENT NAME

- `ECD`

##### SDS VERSION

```
openio 4.7.2.dev18
```


tests/functional/ecd/test_ecd.py
```diff
@@ -0,0 +1,135 @@
...
+                resp = self._download_metachunk(meta, sorted_chunks[pos])
+                self.assertEqual(200, resp.status)
+                data += resp.data
+            self.assertEqual(expected_data, data)
+
+        for pos in range(len(sorted_chunks)):
+            chunk = random.choice(sorted_chunks[pos])
+            sorted_chunks[pos].remove(chunk)
+            resp = self._download_metachunk(meta, sorted_chunks[pos])
+            self.assertEqual(500, resp.status)
```
murlock: It is not related to this PR but IMHO we should use another error code instead of this generic 500

AymericDu: `oio-swift` returns the same error code (500) in this same case.
What error code do you suggest?

tests/functional/ecd/test_ecd.py
```diff
@@ -0,0 +1,177 @@
...
+        chunks, _, _, meta = self.storage.object_create_ext(
+            self.account, container, obj_name=obj, data=expected_data,
+            policy='EC')
+        storage_method = STORAGE_METHODS.load(meta['chunk_method'])
+        sorted_chunks = _sort_chunks(chunks, storage_method.ec)
+
+        def _get_rawx_name(chunk):
+            rawx_addr = chunk['real_url'].split('/')[2]
+            rawx_name = None
+            for rawx in self.conf['services']['rawx']:
```
fvennetier: I suggest using `self.service_to_gridinit_key(urlparse(chunk['url']).netloc), 'rawx')`.

PR #1892, opened 2019-09-19
Improve some cluster commands to accept multiple services

AymericDu:
##### SUMMARY

Improve some cluster commands to accept multiple services

##### ISSUE TYPE

 - Bugfix Pull Request

##### COMPONENT NAME

- `CLI`

##### SDS VERSION

```
openio 4.7.2.dev18
```

### DIFF
--- a/oio/cli/cluster/cluster.py
+++ b/oio/cli/cluster/cluster.py
@@ -14,7 +14,8 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 from logging import getLogger
-from cliff import lister, show
+
+from oio.cli import Lister, ShowOne
 from oio.common.easy_value import boolean_value
 from oio.common.exceptions import OioException, OioNetworkException, \
     ServiceBusy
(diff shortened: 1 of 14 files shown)


oio/cli/cluster/cluster.py
```diff
@@ -161,35 +172,32 @@ def get_parser(self, prog_name):
...
-            yield type_, addr, "unlocked"
-        except Exception as exc:
-            yield type_, addr, str(exc)
+    def _unlock_services(self, parsed_args):
+        srv_definitions = list()
+        for srv_id in parsed_args.srv_ids:
+            srv_definitions.append(
+                self.app.client_manager.cluster.get_service_definition(
+                    parsed_args.srv_type, srv_id))
+        for batch in _bounded_batches(srv_definitions, 4096):
```
fvennetier: 4096 is a bit unusual. Our other paginated requests usually build pages of 1000 elements. It may be a good idea to define that at the top of the file, or as a default value in `_bounded_batches`.

PR #1895, opened 2019-09-23
Allow to set various Redis parameters

AymericDu:
##### SUMMARY

Allow to set various Redis parameters

Jira: OB-461

##### ISSUE TYPE

 - Bugfix Pull Request

##### COMPONENT NAME

- `account`

##### SDS VERSION

```
openio 4.7.2.dev19
```

### DIFF
--- a/oio/common/redis_conn.py
+++ b/oio/common/redis_conn.py
@@ -59,8 +59,10 @@ class RedisConnection(object):
         self._sentinel_hosts = [(h, int(p)) for h, p, in (hp.rsplit(':', 1)
                                 for hp in sentinel_hosts)]
         self._sentinel_name = sentinel_name
+        self._sentinel_conn_kwargs = self._filter_sentinel_conn_kwargs(kwargs)
         self._sentinel = self.__redis_sentinel_mod.Sentinel(
             self._sentinel_hosts,
+            sentinel_kwargs=self._sentinel_conn_kwargs,
             **self._conn_kwargs)
 
     def _filter_conn_kwargs(self, conn_kwargs):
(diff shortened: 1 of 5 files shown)


oio/common/redis_conn.py
```diff
@@ -16,41 +16,103 @@
...
+
+    @property
+    def conn_slave(self):
+        """Retrieve Redis connection (normal or sentinel)"""
+        if self._sentinel:
+            return self._sentinel.master_for(self._sentinel_name)
+        if not self._conn:
+            self._conn = self.__redis_mod.StrictRedis(
+                host=self._host, port=self._port,
+                **self._conn_kwargs)
```
fvennetier: I would return `self.conn` to avoid duplicating these 4 lines.

oio/common/redis_conn.py
```diff
@@ -16,41 +16,99 @@
...
         return self._conn
 
+    @property
+    def conn_slave(self):
+        """Retrieve Redis connection (normal or sentinel)"""
+        if self._sentinel:
+            return self._sentinel.master_for(self._sentinel_name)
```
murlock: IMP it should be `slave_for`

oio/container/backup.py
```diff
@@ -794,7 +795,25 @@ def __init__(self, conf):
         self.REDIS_TIMEOUT = self.conf.get("redis_cache_timeout",
                                            self.REDIS_TIMEOUT)
 
-        super(ContainerBackup, self).__init__(self.conf)
+        redis_conf = {k[6:]: v for k, v in self.conf.items()
```
murlock: Maybe we should factorize code to extract redis paramters

oio/common/redis_conn.py
AymericDu: Oops

PR #1898, opened 2019-09-25
Return 404 when the container created but not initiated

AymericDu:
##### SUMMARY

Return 404 when the container created but not initiated

##### ISSUE TYPE

 - Bugfix Pull Request

##### COMPONENT NAME

- `meta2`

##### SDS VERSION

```
openio 4.7.2.dev23
```

### DIFF
--- a/sqliterepo/replication_dispatcher.c
+++ b/sqliterepo/replication_dispatcher.c
@@ -2013,7 +2013,6 @@ _handler_PROPGET(struct gridd_reply_ctx_s *reply,
 		? (SQLX_OPEN_LOCAL|SQLX_OPEN_NOREFCHECK) : SQLX_OPEN_MASTERSLAVE;
 	err = sqlx_repository_open_and_lock(repo, &n0, how, &sq3, NULL);
 	if (err) {
-		g_prefix_error(&err, "Open/lock: ");
 		reply->send_error(0, err);
 	} else {
 		GPtrArray *tmp = g_ptr_array_new();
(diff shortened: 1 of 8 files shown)


sqliterepo/repository.c
```diff
@@ -481,36 +481,49 @@ sqlx_repository_configure_type(sqlx_repository_t *repo,
 }
 
 void
-sqlx_repository_configure_close_callback(sqlx_repository_t *repo,
-		sqlx_repo_close_hook cb, gpointer cb_data)
+sqlx_repository_configure_db_check_opening_callback(
+		sqlx_repository_t *repo,
+		sqlx_repo_db_check_opening_hook cb, gpointer cb_data)
```
fvennetier: I would have kept `sqlx_repository_configure_open_callback` and `sqlx_repo_open_hook`, with the addition of the open mode.

sqliterepo/replication_dispatcher.c
```diff
@@ -1903,6 +1903,27 @@ _handler_RESYNC(struct gridd_reply_ctx_s *reply,
...
+
+	gchar admin[16];
+	memset(admin, 0, sizeof(admin));
+	err = metautils_message_extract_string(reply->request,
+			NAME_MSGKEY_ADMIN_COMMAND, admin, sizeof(admin));
+	if (err)
+		g_clear_error(&err);
+
+	if (!oio_str_parse_bool(admin, FALSE))
+		err = sqlx_repository_call_db_check_opening_callback(*result, how);
```
fvennetier: The check of the "deleted" flag should be done here, in any case.

PR #1900, opened 2019-09-27
Fix distance computation for chunks selected close to the proxy

fvennetier:
##### SUMMARY
Fix distance computation for chunks selected close to the proxy. It was 0 all the time, triggering an unwanted `storage.content.perfectible` event.

Also bring minor improvements to the CLI.

##### ISSUE TYPE
 - Feature Pull Request
 - Bugfix Pull Request

##### COMPONENT NAME
- LB
- CLI
- sqliterepo

##### SDS VERSION
```
openio 5.2.2.dev10
```


oio/cli/election/election.py
```diff
@@ -117,6 +117,12 @@ def take_action(self, parsed_args):
 class ElectionDebug(ElectionCmd):
     """Get debugging information about an election."""
 
+    def get_parser(self, prog_name):
+        parser = super(ElectionDebug, self).get_parser(prog_name)
+        parser.add_argument('--human', action='store_true',
+                            help="Display human-readable dates")
```
murlock: :+1: Good idea, we should reuse this option

sqliterepo/election.c
```diff
@@ -1352,7 +1352,11 @@ member_json (struct election_member_s *m, GString *gs)
 	g_string_append_c (gs, ',');
 	OIO_JSON_append_str (gs, "type", m->inline_name.type);
 	g_string_append_c (gs, ',');
-	OIO_JSON_append_str (gs, "zk", m->key);
+	gchar *full_key = sqlx_sync_zk_full_key_path(m->sync, m->key);
```
murlock: :+1: 

Maybe we should also show which zk cluster is used for this election ? (however it seems to overlaps `zk_server`)

tests/functional/content/test_content_perfectible.py
```diff
@@ -264,3 +260,18 @@ def test_blob_improver_threecopies(self):
         if job:
             logging.debug("Unexpected job data: %s", data)
         self.assertIsNone(job)
+
+
+class TestPerfectibleLocalContent(TestPerfectibleContent):
+
+    def setUp(self):
```
AymericDu: We could use `setUpClass` and `tearDownClass` to do it once.

oio/cli/election/election.py
fvennetier: Sorry, I moved this to #1904, which is more ZooKeeper themed.

PR #1901, opened 2019-10-01
Decache meta1 database

AymericDu:
##### SUMMARY

Decache when deleting meta1 database or when the base is not managed.

Jira: OS-437

##### ISSUE TYPE

 - Bugfix Pull Request

##### COMPONENT NAME

- `meta1`
- `proxy`

##### SDS VERSION

```
openio 4.7.2.dev41
```

### DIFF
--- a/meta1v2/meta1_backend.c
+++ b/meta1v2/meta1_backend.c
@@ -1,7 +1,7 @@
 /*
 OpenIO SDS meta1v2
 Copyright (C) 2014 Worldline, as part of Redcurrant
-Copyright (C) 2015-2017 OpenIO SAS, as part of OpenIO SDS
+Copyright (C) 2015-2019 OpenIO SAS, as part of OpenIO SDS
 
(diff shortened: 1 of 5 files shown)


meta1v2/meta1_backend.c
```diff
@@ -110,3 +121,27 @@ meta1_backend_get_prefixes(struct meta1_backend_s *m1)
...
+		struct meta1_backend_s *m1b)
+{
+	gint64 seq = 1;
+	EXTRA_ASSERT(sq3 != NULL);
+
+	if (!sq3->deleted)
+		return;
+
+	struct oio_url_s *url = oio_url_empty ();
+	oio_url_set(url, OIOURL_NS, m1b->ns_name);
```
fvennetier: Namespace name is in `sq3->name->ns`. Instead of a `struct meta1_backend_s *`, the second parameter of the callback could be a `struct hc_resolver_s *`, which would avoid adding a field to `struct meta1_backend_s`.

PR #1903, opened 2019-10-03
blob-rebuilder: Fix reuse of beanstalkd connections

AymericDu:
##### SUMMARY

Fix reuse of beanstalkd connections

##### ISSUE TYPE

 - Bugfix Pull Request

##### COMPONENT NAME

- `oio-blob-rebuilder`

##### SDS VERSION

```
openio 4.7.2.dev41
```

### DIFF
--- a/oio/rebuilder/blob_rebuilder.py
+++ b/oio/rebuilder/blob_rebuilder.py
@@ -423,13 +423,19 @@ class BlobRebuilderWorker(RebuilderWorker):
                             reply['addr'], reply['tube'], self.logger,
                             **kwargs)
                     elif self.sender.addr != reply['addr'] \
-                            or self.sender.addr != reply['tube']:
+                            or self.sender.tube != reply['tube']:
                         self.sender.close()
                         self.sender = BeanstalkdSender(
                             reply['addr'], reply['tube'], self.logger,
                             **kwargs)
 
-                    self.sender.send_job(json.dumps(event))
+                    sent = False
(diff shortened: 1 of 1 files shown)


oio/rebuilder/blob_rebuilder.py
```diff
@@ -423,13 +423,19 @@ def update_processed(self, chunk, bytes_processed, error=None, **kwargs):
...
                             reply['addr'], reply['tube'], self.logger,
                             **kwargs)
 
-                    self.sender.send_job(json.dumps(event))
+                    sent = False
+                    event_json = json.dumps(event)
+                    while not sent:
+                        sent = self.sender.send_job(event_json)
+                        if not sent:
+                            sleep(1.0)
```
fvennetier: There is a risk of looping forever here.

AymericDu: If we can not send any more events, we will not have a reply. And if we have no reply for 5 minutes, the program stops (https://github.com/open-io/oio-sds/blob/4.x/oio/rebuilder/blob_rebuilder.py#L311).

fvennetier: No, nothing will timeout here. If for some reason we cannot connect to the beanstalkd server, `send_job` will keep returning `False` (after logging a warning message).

But I'm ok with **not** fixing this: this is an "interactive" program, and the user will see the connection problem (and hopefully fix it).

PR #1906, opened 2019-10-08
rawx: Add hints and optimizations (fallocate(), fsync(), fadvise())

jfsmig:
##### SUMMARY
* Perform still less but bigger I/O ops
* Minor code reorg (most const have been gathered)
* Allow a call to fallocate() before writing
* Disable fsync() by default (similar to rawx-apache2)
* Propose fadvise after uploading of before downloading

##### ISSUE TYPE
enhancement

##### COMPONENT NAME
rawx

##### SDS VERSION
`openio 4.8.1.dev16`


rawx/const.go
```diff
@@ -0,0 +1,115 @@
...
+	uploadBatchSize int = 2048
+
+	// Maximum size (in bytes) of the upload buffer
+	uploadBufferSizeMax int = 8 * 1024 * 1024
+
+	// Minimum size (in bytes) of the upload buffer
+	uploadBufferSizeMin int = 65536
+
+	// Specifies the extension size when Fallocate is called to prepare file placeholders
+	uploadExtensionSize int64 = 64 * 1024 * 1024
```
fvennetier: Shouldn't this be capped to chunk size?

jfsmig: It should, but in the current state of the rawx, the chunksize is not known yet.
And nearly all the deployments tend to prefer a chunk size around 100MiB.

PR #1925, opened 2019-11-17
xcute: Send the tasks by batch

AymericDu:
##### SUMMARY

- Stop using pickle
- Send the tasks by batch
- Allow to use ratelimit
- Count the errors by type
- Use python time instead of redis time
- Keep the job in the workers' cache

##### ISSUE TYPE

 - Feature Pull Request
 - Bugfix Pull Request

##### COMPONENT NAME

- `xcute`

##### SDS VERSION

```
openio 5.2.2
```

### DIFF
--- a/oio/blob/converter.py
+++ b/oio/blob/converter.py
@@ -54,8 +54,8 @@ class BlobConverter(object):
         self.volume = volume
         self.namespace, self.volume_id = check_volume(self.volume)
         # cache
-        self.name_by_cid = CacheDict(size=262144)
-        self.content_id_by_name = CacheDict(size=262144)
+        self.name_by_cid = CacheDict()
+        self.content_id_by_name = CacheDict()
         # client
         self.container_client = ContainerClient(conf, **kwargs)
         self.content_factory = ContentFactory(conf, self.container_client,
--- a/oio/xcute/common/backend.py
+++ b/oio/xcute/common/backend.py
@@ -20,8 +20,7 @@ import random
 
 from oio.common.easy_value import true_value
 from oio.common.exceptions import Forbidden, NotFound
-from oio.common.green import datetime, time
-from oio.common.json import json
+from oio.common.green import datetime
 from oio.common.redis_conn import RedisConnection
 from oio.common.timestamp import Timestamp
 
@@ -45,18 +44,24 @@ class XcuteBackend(RedisConnection):
     DEFAULT_LIMIT = 1000
 
     _lua_errors = {
-        'job_exists': (Forbidden,
-                       'The job already exists'),
-        'no_job': (NotFound,
-                   'The job does\'nt exist'),
-        'lock_exists': (Forbidden,
-                        'The lock already exists'),
-        'must_be_running': (Forbidden,
-                            'The job must be running'),
-        'must_be_paused': (Forbidden,
-                           'The job must be paused'),
-        'must_be_waiting_paused_finished': (Forbidden,
-                                            'The job must be waiting or paused or finished')
+        'job_exists': (
+            Forbidden,
+            'The job already exists'),
+        'no_job': (
+            NotFound,
+            'The job does\'nt exist'),
+        'lock_exists': (
+            Forbidden,
+            'The lock already exists'),
+        'must_be_running': (
+            Forbidden,
+            'The job must be running'),
+        'must_be_paused': (
+            Forbidden,
+            'The job must be paused'),
+        'must_be_waiting_paused_finished': (
+            Forbidden,
+            'The job must be waiting or paused or finished')
         }
 
     key_job_conf = 'xcute:job:config:%s'
@@ -112,7 +117,8 @@ class XcuteBackend(RedisConnection):
                    job_id);
     """ + _lua_update_mtime + """
         local job_info = redis.call('HGETALL', 'xcute:job:info:' .. job_id);
-        local job_config = redis.call('HGETALL', 'xcute:job:config:' .. job_id);
+        local job_config = redis.call(
+            'HGETALL', 'xcute:job:config:' .. job_id);
         return {job_id, job_info, job_config};
     """
 
@@ -299,8 +305,9 @@ class XcuteBackend(RedisConnection):
                         'job.status', 'PAUSED');
                 local orchestrator_id = redis.call(
                     'HGET', 'xcute:job:info:' .. job_id, 'orchestrator.id');
-                redis.call('SREM', 'xcute:orchestrator:jobs:' .. orchestrator_id,
-                        job_id);
+                redis.call(
+                    'SREM', 'xcute:orchestrator:jobs:' .. orchestrator_id,
+                    job_id);
                 stopped = true;
             end;
         end;
@@ -347,7 +354,8 @@ class XcuteBackend(RedisConnection):
         if all_tasks_sent == 'True' and status ~= 'FINISHED' then
             local total_tasks_sent = redis.call(
                 'HGET', 'xcute:job:info:' .. job_id, 'tasks.sent');
-            if tonumber(total_tasks_processed) >= tonumber(total_tasks_sent) then
+            if tonumber(total_tasks_processed) >= tonumber(
+                    total_tasks_sent) then
                 redis.call('HSET', 'xcute:job:info:' .. job_id,
                            'job.status', 'FINISHED');
                 finished = true;
@@ -526,7 +534,8 @@ class XcuteBackend(RedisConnection):
     def update_tasks_sent(self, job_id, task_ids, total_tasks,
                           all_tasks_sent=False):
         return self.script_update_tasks_sent(
-            keys=[self._get_timestamp(), job_id, total_tasks, str(all_tasks_sent)],
+            keys=[self._get_timestamp(), job_id, total_tasks,
+                  str(all_tasks_sent)],
             args=task_ids, client=self.conn)
 
     @handle_redis_exceptions
@@ -543,7 +552,8 @@ class XcuteBackend(RedisConnection):
             for key, value in task_results.items():
                 counters['results.' + key] = value
         return self.script_update_tasks_processed(
-            keys=[self._get_timestamp(), job_id] + self._dict_to_lua_array(counters),
+            keys=[self._get_timestamp(),
+                  job_id] + self._dict_to_lua_array(counters),
             args=task_ids,
             client=self.conn)
 
@@ -592,7 +602,8 @@ class XcuteBackend(RedisConnection):
         job_info = marshalled_job_info.copy()
 
         all_sent = true_value(job_info['tasks.all_sent'])
-        total = int(job_info['tasks.total']) if 'tasks.total' in job_info else None
+        total = int(job_info['tasks.total']) if 'tasks.total' \
+            in job_info else None
 
         job_info['tasks.sent'] = int(job_info['tasks.sent'])
         job_info['tasks.all_sent'] = all_sent
--- a/oio/xcute/common/backend.py
+++ b/oio/xcute/common/backend.py
@@ -516,7 +516,7 @@ class XcuteBackend(RedisConnection):
                                task_errors, task_results):
         counters = dict()
         if task_errors:
-            counters['errors.total'] = len(task_errors)
+            counters['errors.total'] = task_errors
         if task_results:
             for key, value in task_results.items():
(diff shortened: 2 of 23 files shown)


oio/blob/converter.py
```diff
@@ -54,8 +54,8 @@ def __init__(self, conf, logger=None, **kwargs):
         self.volume = volume
         self.namespace, self.volume_id = check_volume(self.volume)
         # cache
-        self.name_by_cid = CacheDict(size=262144)
-        self.content_id_by_name = CacheDict(size=262144)
+        self.name_by_cid = CacheDict()
```
loamhoof: Shouldn't this be in another PR, not in xcute?

oio/xcute/common/backend.py
```diff
@@ -367,12 +379,13 @@ class XcuteBackend(RedisConnection):
...
 
         if status == 'WAITING' then
-            redis.call('LREM', 'xcute:waiting:jobs', 1, KEYS[1]);
+            redis.call('LREM', 'xcute:waiting:jobs', 1, job_id);
         end;
 
         redis.call('ZREM', 'xcute:job:ids', job_id);
         redis.call('DEL', 'xcute:job:info:' .. job_id);
         redis.call('DEL', 'xcute:job:config:' .. job_id);
+        redis.call('DEL', 'xcute:tasks:running:' .. job_id);
```
loamhoof: This should have been in another commit.

oio/xcute/common/job.py
```diff
@@ -17,9 +17,22 @@
 from oio.common.logger import get_logger
 
 
+class XcuteTask(object):
+
+    def __init__(self, conf, job_config, logger=None):
+        self.conf = conf
+        self.logger = logger or get_logger(self.conf)
+
+        self.tasks_per_second = int(job_config['tasks_per_second'])
```
loamhoof: Why does the task need to know that?
Also why does it need to be cast?

oio/xcute/common/job.py
```diff
@@ -17,20 +17,33 @@
...
+
+
 class XcuteJob(object):
 
     JOB_TYPE = None
+    TASK_CLASS = None
 
     DEFAULT_TASKS_PER_SECOND = 32
-    MAX_TASKS_BATCH_SIZE = 256
+    MAX_TASKS_BATCH_SIZE = 512
```
loamhoof: That's still completely arbitrary.

oio/xcute/common/job.py
```diff
@@ -44,12 +57,15 @@ def load_config(self, job_config):
         self.tasks_batch_size = int_value(
             job_config.get('tasks_batch_size'), None)
         if self.tasks_batch_size is None:
-            self.tasks_batch_size = min(
-                self.tasks_per_second * 2, self.MAX_TASKS_BATCH_SIZE)
+            if self.tasks_per_second > 0:
+                self.tasks_batch_size = min(
+                    self.tasks_per_second * 2, self.MAX_TASKS_BATCH_SIZE)
+            else:
+                self.tasks_batch_size = self.MAX_TASKS_BATCH_SIZE
```
loamhoof: Why is `sanitize_params` still stateful?

oio/xcute/common/worker.py
```diff
@@ -29,22 +29,19 @@ def __init__(self, conf, logger=None):
...
-            job_class = JOB_TYPES[job_type]
+            task_class = JOB_TYPES[job_type].TASK_CLASS
             job_config = beanstalkd_job['job_config']
 
-            job = job_class(self.conf, logger=self.logger)
-            job.load_config(job_config)
-            job.init_process_task()
-
-            self.jobs[job_id] = job
+            task = task_class(self.conf, job_config, logger=self.logger)
```
loamhoof: I still think the task doesn't need to be passed all of the job config, but only its specific parameters.

oio/xcute/jobs/mover.py
```diff
@@ -21,27 +21,92 @@
...
+
+        # Start moving the chunk
+        try:
+            content = self.content_factory.get(container_id, content_id)
+            content.move_chunk(
+                chunk_id, fake_excluded_chunks=self.fake_excluded_chunks)
+        except (ContentNotFound, OrphanChunk):
+            return {'orphan': 1}
+
+        return {'moved_chunks': 1, 'total_size': chunk_size}
```
loamhoof: Lacking some consistency with the names: `skipped`, `orphan` but `moved_chunks`?

oio/xcute/orchestrator.py
```diff
@@ -135,7 +135,7 @@ def handle_new_job(self, job_id, job_type, last_task_id, job_config):
 
         job_class = JOB_TYPES[job_type]
         job = job_class(self.conf, logger=self.logger)
-        job.load_config(job_config)
+        job.sanitize_params(job_config)
```
loamhoof: Why is this here? It should only be called in the server.

oio/xcute/common/job.py
```diff
@@ -57,8 +57,11 @@ def sanitize_params(self, job_config):
         self.tasks_batch_size = int_value(
             job_config.get('tasks_batch_size'), None)
         if self.tasks_batch_size is None:
-            self.tasks_batch_size = min(
-                self.tasks_per_second * 2, self.MAX_TASKS_BATCH_SIZE)
+            if self.tasks_per_second > 0:
+                self.tasks_batch_size = min(
+                    self.tasks_per_second * 2, self.MAX_TASKS_BATCH_SIZE)
+            else:
```
loamhoof: That should return an error?

oio/xcute/orchestrator.py
```diff
@@ -118,127 +117,148 @@ def orchestrate_loop(self):
...
             Dispatch all of a job's tasks
         """
 
         self.logger.info('Start dispatching job (job_id=%s)', job_id)
 
         try:
-            task = None
-            for task in job_tasks:
-                (task_class, task_id, task_payload, total_tasks) = task
+            items_run_time = 0
```
loamhoof: What is `items`?

oio/xcute/common/job.py
AymericDu: When `tasks_per_second` is less than or equal to 0, it means we want to go as fast as possible.

oio/xcute/common/job.py
```diff
@@ -58,13 +71,7 @@ def load_config(self, job_config):
     def get_tasks(self, marker=None):
         """
             Yields the job tasks as
-            (TaskClass, task_id, task_payload, total_tasks)
+            (task_id, task_payload, total_tasks)
```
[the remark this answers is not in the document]
AymericDu: In which case would a job want to send different tasks? And even if it were, it would be enough to send the information to the payload so that the task processing uses the correct code.

oio/xcute/common/job.py
AymericDu: So what value do you propose? This value depends on the size of the task IDs and their payloads. These two sizes depending on the type of job. It is therefore impossible to have a default value that will be perfect for all jobs. If a job wants to change this value, it can override it in its class.

oio/xcute/common/backend.py
```diff
@@ -471,58 +490,73 @@ def list_orchestrator_jobs(self, orchestrator_id):
...
         if next_job is None:
             return None
         job_id, job_info, job_config = next_job
         job_info = self._unmarshal_job_info(
             self._lua_array_to_dict(job_info))
-        job_config = self._unmarshal_job_conf(
-            self._lua_array_to_dict(job_config))
-        return (job_id, job_info['job_type'], job_info.get('last_sent'),
-                job_config)
+        job_config = self._lua_array_to_dict(job_config)
```
loamhoof: Consistency between `job_info` and `job_config`

oio/xcute/common/job.py
loamhoof: If you want a default and not do it per class, imho:
- it should be commented to say why we have one, and the consequences
- isn't it too big? 512 allows <128 characters by task only

oio/xcute/common/job.py
```diff
@@ -30,23 +30,63 @@ def process(self, task_id, task_payload):
...
 
     JOB_TYPE = None
+    TASK_CLASS = None
+
+    DEFAULT_TASKS_PER_SECOND = 32
+    MAX_TASKS_BATCH_SIZE = 512
 
-    @staticmethod
-    def sanitize_params(params):
+    def __init__(self, conf, logger=None):
```
loamhoof: Please go back to the static version. There is no reason to have a state

oio/xcute/common/job.py
```diff
@@ -30,23 +30,63 @@ def process(self, task_id, task_payload):
...
+        """
+        sanitized_job_config = dict()
+
+        tasks_per_second = int_value(
+            job_config.get('tasks_per_second'),
+            self.DEFAULT_TASKS_PER_SECOND)
+        sanitized_job_config['tasks_per_second'] = tasks_per_second
+
+        tasks_batch_size = int_value(
+            job_config.get('tasks_batch_size'), None)
```
loamhoof: `.get(foo, None)` is redudant

oio/xcute/common/job.py
```diff
@@ -30,23 +30,63 @@ def process(self, task_id, task_payload):
...
+
+        return sanitized_job_config, lock
+
+    def sanitize_params(self, job_params):
         """
             Validate and sanitize the job parameters
             Ex: cast a string as integer, set a default
             Also return the lock id if there is one
         """
+        sanitized_job_params = dict()
```
loamhoof: Why not raise anymore?

oio/xcute/common/job.py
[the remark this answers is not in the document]
loamhoof: So the only reason we were sending the task at the beginning was because we wanted to pickle it? Ok

oio/xcute/common/backend.py
```diff
@@ -471,58 +490,73 @@ def list_orchestrator_jobs(self, orchestrator_id):
...
-            keys=[job_id, total_tasks, str(all_tasks_sent)],
+            keys=[self._get_timestamp(), job_id, total_tasks,
+                  str(all_tasks_sent)],
             args=task_ids, client=self.conn)
 
     @handle_redis_exceptions
     def update_tasks_processed(self, job_id, task_ids,
                                task_errors, task_results):
         counters = dict()
         if task_errors:
```
loamhoof: `task_errors` & `task_results` should never be empty. If we wanna double make sure they're not empty, get them with a `.get('foo', {})` in the orchestrator. That would remove two `ifs`

oio/xcute/common/job.py
```diff
@@ -30,23 +30,63 @@ def process(self, task_id, task_payload):
...
+            job_config.get('tasks_per_second'),
+            self.DEFAULT_TASKS_PER_SECOND)
+        sanitized_job_config['tasks_per_second'] = tasks_per_second
+
+        tasks_batch_size = int_value(
+            job_config.get('tasks_batch_size'), None)
+        if tasks_batch_size is None:
+            if tasks_per_second > 0:
+                tasks_batch_size = min(
+                    tasks_per_second * 2, self.MAX_TASKS_BATCH_SIZE)
```
loamhoof: It'd be nice to have a default batch size "optimised" for every job. A 2 seconds worth of tasks batch size is huge.

oio/xcute/common/job.py
AymericDu: `None` is not `get` parameter, but` int_value` parameter.

oio/xcute/common/job.py
```diff
@@ -30,23 +30,63 @@ def process(self, task_id, task_payload):
...
+                tasks_batch_size = self.MAX_TASKS_BATCH_SIZE
+        elif tasks_batch_size < 1:
+            raise ValueError('Tasks batch size should be positive')
+        elif tasks_batch_size > self.MAX_TASKS_BATCH_SIZE:
+            raise ValueError('Tasks batch size should be less than %d' %
+                             self.MAX_TASKS_BATCH_SIZE)
+        sanitized_job_config['tasks_batch_size'] = tasks_batch_size
+
+        sanitized_job_params, lock = self.sanitize_params(
+            job_config.get('params') or dict())
```
loamhoof: `.get('params', dict())`

AymericDu: If the `params` key is in `job_config`, but its value is `None`, we must be sure to have a `dict`.

PR #1927, opened 2019-11-18
Fix ec chunk handling by not trying to recover from the stream

newtoncorp:
##### SUMMARY
This patch fixes the EC path in case an exception is raised while downloading
a chunk. Previously the code would augment the range. This code is probably
intended for three copy code path but break EC. Set a flag and check it
before trying to 'recover'.

##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT NAME
OIO

##### SDS VERSION
```
all
```

##### ADDITIONAL INFORMATION
You can reproduce by using tc on a rawx with erasure coded segments

### DIFF
--- a/oio/api/ec.py
+++ b/oio/api/ec.py
@@ -162,7 +162,8 @@ class ECChunkDownloadHandler(object):
                                 headers, self.connection_timeout,
                                 self.read_timeout, perfdata=self.perfdata,
                                 align=True, logger=self.logger,
-                                resp_by_chunk=self._resp_by_chunk)
+                                resp_by_chunk=self._resp_by_chunk,
+                                is_ec=True)
         return (reader, reader.get_iter())
 
     def get_stream(self):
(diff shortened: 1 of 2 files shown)


fvennetier:
Hello. The range change is a feature, not a bug. There are several "fragments" in a chunk, and the reads have to be aligned on theses. When a download fails in the middle, we must start to read the backup chunk at the start of the corresponding fragment.

Here I implemented a test simulating failure in the middle of the download of a chunk: https://github.com/fvennetier/oio-sds/commit/cd5ec3a0f2805a73b1378156003c1c54eda46483
If we apply this pull request, the test won't pass.

newtoncorp:
The range recover will make the erasure code fail if it breaks in the middle of a segment.
Try doing a test where the second segments fails at offset 42 on a 6+3 configuration with an ec segment available.

Something like :
 - seg1 -> ok
 - seg2 -> ok
 - seg3 -> ok
 - seg4 -> short read at offset 42 (as long as it's not a chunk boundary)
 - seg5 -> ok
 - seg6 -> ok
 - seg7 -> ok
 - seg8 -> not used
 - seg9 -> not used

newtoncorp:
😞

PR #1929, opened 2019-11-21
xcute: Implement blob rebuilder

AymericDu:
##### SUMMARY

Implement blob rebuilder

##### ISSUE TYPE

 - Feature Pull Request

##### COMPONENT NAME

- `xcute`

##### SDS VERSION

```
openio 5.2.2
```

### DIFF
--- a/oio/cli/cluster/cluster.py
+++ b/oio/cli/cluster/cluster.py
@@ -93,7 +93,7 @@ class ClusterList(Lister):
                 volume = tags.get('tag.vol', 'n/a')
                 service_id = tags.get('tag.service_id', 'n/a')
                 addr = srv['addr']
-                locked = boolean_value(tags.get('tag.lock', False))
+                locked = boolean_value(tags.get('tag.lock'), False)
                 up = tags.get('tag.up', 'n/a')
                 score = srv['score']
                 if parsed_args.stats:
@@ -136,8 +136,8 @@ class ClusterLocalList(Lister):
         results = []
         srv_types = parsed_args.srv_types
         local_scores = boolean_value(
-            self.app.client_manager.sds_conf.get('proxy.quirk.local_scores',
-                                                 False))
+            self.app.client_manager.sds_conf.get('proxy.quirk.local_scores'),
+            False)
         if not local_scores:
             self.log.warn("'proxy.quirk.local_scores' not set, "
                           "scores won't be realistic.")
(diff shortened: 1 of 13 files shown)


oio/cli/cluster/cluster.py
```diff
@@ -93,7 +93,7 @@ def _list_services(self, parsed_args):
                 volume = tags.get('tag.vol', 'n/a')
                 service_id = tags.get('tag.service_id', 'n/a')
                 addr = srv['addr']
-                locked = boolean_value(tags.get('tag.lock', False))
+                locked = boolean_value(tags.get('tag.lock'), False)
```
loamhoof: Once again this has nothing to do in xcute?

oio/xcute/server.py
```diff
@@ -73,10 +73,9 @@ def on_job_list(self, req):
             if job_class is None:
                 return HTTPBadRequest(UnknownJobTypeException.message)
 
-            job = job_class(self.conf, logger=self.logger)
-            job_config = data.get('config')
             # TODO: use lock
-            job_config, lock = job.sanitize_config(job_config)
+            job_config, lock = job_class.sanitize_config(
+                data.get('config') or dict())
```
loamhoof: `data.get('config', dict)`?

oio/xcute/common/job.py
```diff
@@ -39,7 +39,8 @@ def __init__(self, conf, logger=None):
         self.conf = conf
         self.logger = logger or get_logger(self.conf)
 
-    def sanitize_config(self, job_config):
+    @classmethod
+    def sanitize_config(cls, job_config):
```
loamhoof: Can we change back `get_tasks` to be static as well?

oio/xcute/server.py
AymericDu:
```
>>> mydict = dict(mykey=None)
>>> mydict.get('mykey')
>>> mydict.get('mykey', dict())
>>> mydict.get('mykey') or dict()
{}
```

loamhoof: Why would you have None though?

loamhoof: If you can have None, you may as well check that you have the right type

oio/xcute/common/job.py
AymericDu: In all our code, we give the configuration and the logger in the constructor. So, I prefer to keep this same logic.

PR #1930, opened 2019-11-21
xcute: Improve loadbalancing strategy

AymericDu:
##### SUMMARY

Improve loadbalancing strategy

##### ISSUE TYPE

 - Feature Pull Request

##### COMPONENT NAME

- xcute

##### SDS VERSION

```
openio 5.2.2
```

### DIFF
--- a/oio/xcute/orchestrator.py
+++ b/oio/xcute/orchestrator.py
@@ -32,6 +32,7 @@ class XcuteOrchestrator(object):
 
     DEFAULT_DISPATCHER_TIMEOUT = 2
     DEFAULT_REFRESH_TIME_BEANSTALKD_WORKERS = 30
+    DEFAULT_MAX_JOBS_PER_BEANSTALKD = 1024
 
     def __init__(self, conf, logger=None):
         self.conf = conf
@@ -62,6 +63,10 @@ class XcuteOrchestrator(object):
         self.refresh_time_beanstalkd_workers = int_value(
             self.conf.get('refresh_time_beanstalkd_workers'),
             self.DEFAULT_REFRESH_TIME_BEANSTALKD_WORKERS)
+
+        self.max_jobs_per_beanstalkd = int_value(
+            self.conf.get('max_jobs_per_beanstalkd'),
+            self.DEFAULT_MAX_JOBS_PER_BEANSTALKD)
 
         self.running = True
         self.beanstalkd_workers = dict()
@@ -452,12 +457,24 @@ class XcuteOrchestrator(object):
 
         current_stats = beanstalkd.stats_tube(
             self.beanstalkd_workers_tube)
-        if current_stats['current-jobs-reserved'] <= 0 \
-                and current_stats['current-jobs-ready'] > 0:
-            self.logger.info(
-                'Ignore beanstalkd %s: The worker doesn\'t process task',
-                beanstalkd_addr)
-            return None
+        beanstalkd_jobs_ready = current_stats['current-jobs-ready']
+        if beanstalkd_jobs_ready > 0:
+            beanstalkd_jobs_reserved = current_stats['current-jobs-reserved']
+            if beanstalkd_jobs_reserved <= 0:
+                self.logger.info(
+                    'Ignore beanstalkd %s: The worker doesn\'t process task '
+                    '(current-jobs-ready=%d, current-jobs-reserved=%d)',
+                    beanstalkd_addr, beanstalkd_jobs_ready,
+                    beanstalkd_jobs_reserved)
+                return None
+
+            if beanstalkd_jobs_ready >= self.max_jobs_per_beanstalkd:
+                self.logger.info(
+                    'Ignore beanstalkd %s: The queue is full '
+                    '(current-jobs-ready=%d, current-jobs-reserved=%d)',
+                    beanstalkd_addr, beanstalkd_jobs_ready,
+                    beanstalkd_jobs_reserved)
+                return None
 
         if hasattr(beanstalkd, 'is_broken') and beanstalkd.is_broken:
             self.logger.info(
--- a/oio/xcute/orchestrator.py
+++ b/oio/xcute/orchestrator.py
@@ -13,6 +13,7 @@
 # You should have received a copy of the GNU Lesser General Public
 # License along with this library.
 
+import random
 from collections import OrderedDict
 
 from oio.common.exceptions import OioTimeout
@@ -21,7 +22,7 @@ from oio.common.green import ratelimit, sleep, threading
 from oio.common.json import json
 from oio.conscience.client import ConscienceClient
 from oio.event.beanstalk import Beanstalk, BeanstalkdListener, \
-    BeanstalkdSender, ConnectionError
+    ConnectionError
 from oio.xcute.common.backend import XcuteBackend
 from oio.xcute.jobs import JOB_TYPES
 
@@ -29,6 +30,7 @@ from oio.xcute.jobs import JOB_TYPES
 class XcuteOrchestrator(object):
 
     DEFAULT_DISPATCHER_TIMEOUT = 2
+    DEFAULT_REFRESH_TIME_BEANSTALKD = 5
 
     def __init__(self, conf, logger=None):
         self.conf = conf
@@ -57,7 +59,8 @@ class XcuteOrchestrator(object):
                          self.beanstalkd_reply_tube)
 
         self.running = True
-        self.threads = {}
+        self.beanstalkd_workers = dict()
+        self.threads = dict()
 
     def run_forever(self):
         """
@@ -65,20 +68,11 @@ class XcuteOrchestrator(object):
         """
 
         # gather beanstalkd info
-        self.all_beanstalkd = OrderedDict()
-        self.beanstalkd_senders = {}
-        beanstalkd_thread = threading.Thread(
-            target=self.refresh_all_beanstalkd)
-        beanstalkd_thread.start()
-
-        self.logger.info('Wait until beanstalkd are found')
-        while len(self.all_beanstalkd) == 0:
-            if not self.running:
-                return
-
-            sleep(5)
-
-        self.threads[beanstalkd_thread.ident] = beanstalkd_thread
+        refresh_beanstalkd_workers_thread = threading.Thread(
+            target=self.refresh_beanstalkd_workers_forever)
+        refresh_beanstalkd_workers_thread.start()
+        self.threads[refresh_beanstalkd_workers_thread.ident] = \
+            refresh_beanstalkd_workers_thread
 
         # restart running jobs
         self.logger.debug('Look for unfinished jobs')
@@ -151,10 +145,7 @@ class XcuteOrchestrator(object):
             tasks_counter = job.get_total_tasks(
                 job_params, marker=total_marker)
 
-        beanstalkd_workers = self.get_loadbalanced_workers()
-
-        thread_args = (job_id, job_type, job_config, job_tasks,
-                       beanstalkd_workers)
+        thread_args = (job_id, job_type, job_config, job_tasks)
         dispatch_thread = threading.Thread(
             target=self.dispatch_job,
             args=thread_args)
@@ -170,8 +161,7 @@ class XcuteOrchestrator(object):
             args=(job_id, tasks_counter))
         total_tasks_thread.start()
 
-    def dispatch_job(self, job_id, job_type, job_config, job_tasks,
-                     beanstalkd_workers):
+    def dispatch_job(self, job_id, job_type, job_config, job_tasks):
         """
             Dispatch all of a job's tasks
         """
@@ -179,6 +169,8 @@ class XcuteOrchestrator(object):
         self.logger.info('Start dispatching job (job_id=%s)', job_id)
 
         try:
+            beanstalkd_workers = self.get_beanstalkd_workers()
+
             tasks_per_second = job_config['tasks_per_second']
             tasks_batch_size = job_config['tasks_batch_size']
 
@@ -246,31 +238,23 @@ class XcuteOrchestrator(object):
                              len(beanstalkd_payload))
 
         while self.running:
-            workers_tried = set()
-            for worker in beanstalkd_workers:
-                if worker is None:
-                    self.logger.info('No beanstalkd available (job_id=%s)',
-                                     job_id)
+            for beanstalkd_worker in beanstalkd_workers:
+                if not self.running:
+                    return False
+                if beanstalkd_worker is not None:
                     break
 
-                if worker.addr in workers_tried:
-                    self.logger.debug('Tried all beanstalkd (job_id=%s)',
-                                      job_id)
-                    break
-
-                sent = worker.send_job(beanstalkd_payload)
-
-                if not sent:
-                    workers_tried.add(worker.addr)
-
-                    continue
-
-                self.logger.debug('Tasks (job_id=%s) sent to %s: %s',
-                                  job_id, worker.addr, tasks)
+            try:
+                beanstalkd_worker.put(beanstalkd_payload)
+                self.logger.debug(
+                    '[job_id=%s] Tasks sent to %s: %s', job_id,
+                    beanstalkd_worker.addr, str(tasks))
                 return True
-
-            workers_tried.clear()
-            sleep(5)
+            except Exception as exc:
+                self.logger.warn(
+                    '[job_id=%s] Fail to send beanstalkd job: %s',
+                    job_id, exc)
+        return False
 
     def make_beanstalkd_payload(self, job_id, job_type, job_config,
                                 tasks):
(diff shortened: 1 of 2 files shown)


oio/xcute/orchestrator.py
```diff
@@ -248,32 +235,29 @@ def dispatch_tasks(self, beanstalkd_workers,
...
+            try:
+                beanstalkd_worker.put(beanstalkd_payload)
+                sent = True
+                break
+            except Exception as exc:
+                self._refresh_beanstalkd_workers = True
+                self.logger.warn(
+                    '[job_id=%s] Fail to send beanstalkd job: %s',
+                    job_id, exc)
+                sleep(1)
```
loamhoof: Why wait when you could try the next one directly?

oio/xcute/orchestrator.py
```diff
@@ -58,28 +59,21 @@ def __init__(self, conf, verbose):
                          self.beanstalkd_reply_tube)
 
         self.running = True
-        self.threads = {}
+        self.beanstalkd_workers = None
+        self.beanstalkd_workers_next_index = 0
```
loamhoof: Can't we make `beanstalkd_workers` an iterator instead of this random global state?

oio/xcute/orchestrator.py
```diff
@@ -248,32 +235,29 @@ def dispatch_tasks(self, beanstalkd_workers,
...
 
-                sent = worker.send_job(beanstalkd_payload)
-
-                if not sent:
-                    workers_tried.add(worker.addr)
-
-                    continue
-
-                self.logger.debug('Tasks (job_id=%s) sent to %s: %s',
-                                  job_id, worker.addr, tasks)
```
loamhoof: Removed the log? :crying_cat_face:

oio/xcute/orchestrator.py
```diff
@@ -366,77 +350,104 @@ def process_reply(self, beanstalkd_job_id, encoded_reply):
...
+                                              beanstalkd_stats)
+            except Exception as exc:
+                self.logger.error(
+                    'Fail to find beanstalkd workers: %s', exc)
 
-            for beanstalkd_addr in self.all_beanstalkd:
-                if beanstalkd_addr in all_beanstalkd_with_tubes:
-                    continue
+            # Shuffle to avoid having the beanstalkd consecutively
+            random.shuffle(beanstalkd_workers)
```
loamhoof: Not sure I understand the point

oio/xcute/orchestrator.py
```diff
@@ -366,77 +350,104 @@ def process_reply(self, beanstalkd_job_id, encoded_reply):
...
-                try:
-                    beanstalkd_tubes = self.get_beanstalkd_tubes(
-                        beanstalkd_addr)
-                except ConnectionError:
-                    continue
-
-                all_beanstalkd_with_tubes[beanstalkd_addr] = (
-                    beanstalkd, beanstalkd_tubes)
+            try:
+                self._find_beanstalkd_workers(beanstalkd_workers,
```
loamhoof: please return `beanstalkd_workers` and `beanstalkd_stats` so it's easier to see that they've been updated

oio/xcute/orchestrator.py
```diff
@@ -366,77 +350,104 @@ def process_reply(self, beanstalkd_job_id, encoded_reply):
...
-                        addr=beanstalkd['addr'],
-                        tube=self.beanstalkd_workers_tube,
-                        logger=self.logger)
+        current_stats = beanstalkd.stats_tube(
+            self.beanstalkd_workers_tube)
+        previous_stats = beanstalkd_stats.get(beanstalkd_addr)
+        beanstalkd_stats[beanstalkd_addr] = current_stats
+        if (previous_stats is not None
+                and (previous_stats['current-jobs-ready'] > 0
+                     or current_stats['current-jobs-ready'] > 0)
```
loamhoof: Is `current-jobs-ready` the number of jobs in the queue?
It's completely normal to have some waiting in the queue (it limits the overhead) and we should strive to always have some in the queue.

oio/xcute/orchestrator.py
```diff
@@ -366,77 +350,104 @@ def process_reply(self, beanstalkd_job_id, encoded_reply):
...
+                and (previous_stats['cmd-delete']
+                     == current_stats['cmd-delete'])):
```
loamhoof: That could work if this function is run every 5 minutes, but yours can be run every second. So if a worker takes more than one second to complete a batch, it's removed for up to 5 minutes?

oio/xcute/orchestrator.py
loamhoof: also why pass an empty list that should be filled?

oio/xcute/orchestrator.py
loamhoof: Ok I see...

oio/xcute/orchestrator.py
```diff
@@ -366,77 +350,104 @@ def process_reply(self, beanstalkd_job_id, encoded_reply):
...
+            # The worker doesn't process task or it's very slow
+            return
+
+        self.logger.info('Found beanstalkd %s', beanstalkd_addr)
+        beanstalkd.use(self.beanstalkd_workers_tube)
+        beanstalkd.watch(self.beanstalkd_workers_tube)
 
-                    beanstalkd_senders[self.beanstalkd_workers_tube] = sender
+        # Favor the beanstalkd with a good score
+        for _ in range(beanstalkd_score/10 or 1):
```
loamhoof: Can we use floats and use `math.ceil` instead of a weird-ass `or 1` that actually breaks the logic?

loamhoof: So we're gonna take an instant value of the score to decide the loadbalancing strategy for up to 5 minutes?
Also, do we really want to send 10 jobs to a worker with a score of 100, when the one with a score of 10 receives only one?

oio/xcute/orchestrator.py
```diff
@@ -366,77 +350,104 @@ def process_reply(self, beanstalkd_job_id, encoded_reply):
...
+    def get_beanstalkd_worker(self):
+        """
+            Get beanstalkd worker following a loadbalancing strategy
+        """
 
-            if not yielded:
-                yield None
+        if not self.beanstalkd_workers:
+            return None
+        next_index = self.beanstalkd_workers_next_index \
```
loamhoof: This is broken when the list dwindles.

oio/xcute/orchestrator.py
AymericDu: Yes, we want that. By cons, I reduced the refresh time.

oio/xcute/orchestrator.py
```diff
@@ -245,32 +248,29 @@ def dispatch_tasks(self, beanstalkd_workers,
...
+                self.logger.debug(
+                    '[job_id=%s] Tasks sent to %s: %s', job_id,
+                    beanstalkd_worker.addr, str(tasks))
+                break
+            except Exception as exc:
+                self.logger.warn(
+                    '[job_id=%s] Fail to send beanstalkd job: %s',
+                    job_id, exc)
+                beanstalkd_worker.is_broken = True
+                continue
```
loamhoof: What does the `continue` do here?

oio/xcute/orchestrator.py
```diff
@@ -245,32 +248,29 @@ def dispatch_tasks(self, beanstalkd_workers,
...
+                sent = True
+                self.logger.debug(
+                    '[job_id=%s] Tasks sent to %s: %s', job_id,
+                    beanstalkd_worker.addr, str(tasks))
+                break
+            except Exception as exc:
+                self.logger.warn(
+                    '[job_id=%s] Fail to send beanstalkd job: %s',
+                    job_id, exc)
+                beanstalkd_worker.is_broken = True
```
loamhoof: Maybe be a little more lenient here, just like it's usually done for heartbeats: count the number of failed requests in a row, and if it's over a certain threshold, consider it as broken.

oio/xcute/orchestrator.py
```diff
@@ -245,32 +248,29 @@ def dispatch_tasks(self, beanstalkd_workers,
...
-                    workers_tried.add(worker.addr)
-
-                    continue
+            try:
+                beanstalkd_worker.put(beanstalkd_payload)
+                sent = True
+                self.logger.debug(
+                    '[job_id=%s] Tasks sent to %s: %s', job_id,
+                    beanstalkd_worker.addr, str(tasks))
+                break
```
loamhoof: Let's just return directly, we're not doing anything after that, right?

oio/xcute/orchestrator.py
```diff
@@ -375,77 +375,144 @@ def process_reply(self, beanstalkd_job_id, encoded_reply):
...
-                    continue
-
-                all_beanstalkd_with_tubes[beanstalkd_addr] = (
-                    beanstalkd, beanstalkd_tubes)
+            try:
+                beanstalkd_workers = self._find_beanstalkd_workers()
+            except Exception as exc:
+                self.logger.error(
+                    'Fail to find beanstalkd workers: %s', exc)
+                beanstalkd_workers = dict()
```
loamhoof: Isn't it a bit too much? We could keep trying to send jobs to the beanstalkd we already found. If there's really a problem, they will be marked as broken anyway.

oio/xcute/orchestrator.py
```diff
@@ -375,77 +375,144 @@ def process_reply(self, beanstalkd_job_id, encoded_reply):
...
 
-            for beanstalkd_addr in self.all_beanstalkd:
-                if beanstalkd_addr in all_beanstalkd_with_tubes:
-                    continue
+            removed_beanstalkds = set(self.beanstalkd_workers.keys()) \
+                - set(beanstalkd_workers.keys())
+            for beanstalkd_addr in removed_beanstalkds:
+                self.logger.info('Remove beanstalkd %s' % beanstalkd_addr)
```
loamhoof: That would also be nice to have a log when we have a new one.

oio/xcute/orchestrator.py
```diff
@@ -375,77 +375,144 @@ def process_reply(self, beanstalkd_job_id, encoded_reply):
...
-                self.logger.info('Removed beanstalkd %s' % beanstalkd_addr)
-                del self.all_beanstalkd[beanstalkd_addr]
+            # Refresh beanstalkd workers
+            self.beanstalkd_workers = beanstalkd_workers
 
-            for beanstalkd_addr, beanstalkd \
-                    in all_beanstalkd_with_tubes.iteritems():
-                if beanstalkd_addr not in self.all_beanstalkd:
-                    self.logger.info('Found beanstalkd %s' % beanstalkd_addr)
+            for i in range(self.refresh_time_beanstalkd_workers):
```
loamhoof: `i` unused, `_` would be better? Or does the code style prefer this version?

oio/xcute/orchestrator.py
loamhoof: Nvm, it's down under!

oio/xcute/orchestrator.py
```diff
@@ -375,77 +375,144 @@ def process_reply(self, beanstalkd_job_id, encoded_reply):
...
+            new_beanstalkd = True
+            beanstalkd = Beanstalk.from_url(beanstalkd_addr)
+            beanstalkd.addr = beanstalkd_addr
+        beanstalkd.score = beanstalkd_score
+
+        beanstalkd_tubes = beanstalkd.tubes()
+        if self.beanstalkd_workers_tube not in beanstalkd_tubes:
+            self.logger.info(
+                'Ignore beanstalkd %s: '
+                'No worker has ever listened to the tube', beanstalkd_addr)
```
loamhoof: Add the tube name to the log maybe

oio/xcute/orchestrator.py
```diff
@@ -375,77 +375,144 @@ def process_reply(self, beanstalkd_job_id, encoded_reply):
...
+            return None
+
+        current_stats = beanstalkd.stats_tube(
+            self.beanstalkd_workers_tube)
+        beanstalkd_jobs_ready = current_stats['current-jobs-ready']
+        if beanstalkd_jobs_ready > 0:
+            if current_stats['current-jobs-reserved'] <= 0:
```
loamhoof: Can't we have this condition randomly happening?

loamhoof: Can't this happen if the worker has just finished a job and hasn't taken another one yet?

oio/xcute/orchestrator.py
```diff
@@ -375,77 +375,144 @@ def process_reply(self, beanstalkd_job_id, encoded_reply):
...
-                        tube=self.beanstalkd_workers_tube,
-                        logger=self.logger)
+            beanstalkd_workers_id = id(self.beanstalkd_workers)
 
-                    beanstalkd_senders[self.beanstalkd_workers_tube] = sender
+            # Favor the beanstalkd with a good score
+            beanstalkd_workers = list()
+            for beanstalkd in self.beanstalkd_workers.values():
+                occurrence = int(math.ceil(beanstalkd.score/10.))
+                occurrence += random.randint(-2, 2)
```
loamhoof: Is it a formula we're already using? A known loadbalancing method?

oio/xcute/orchestrator.py
AymericDu: If there was an error, it is surely because of a problem. So let's leave it alone until the next refreshment which is coming soon (30 seconds).

oio/xcute/orchestrator.py
AymericDu: The only errors that can occur here are conscience errors. If there is a problem with the conscience, it is better to wait until the problem is resolved before for the good progress of the job.

oio/xcute/orchestrator.py
AymericDu: In this case, we will (wrongly) ignore this beanstalkd. But this case has a fairly low probability of happening with the configuration of the event agent for xcute (workers = 2, concurrency = 5).

oio/xcute/orchestrator.py
loamhoof: Why? We don't need the conscience to complete the job, do we?

oio/xcute/orchestrator.py
loamhoof: Right, and there's a log at least, so I'm ok :+1:
Maybe just add in the log the jobs-ready, jobs-reserved counts though!

oio/xcute/orchestrator.py
```diff
@@ -375,77 +371,154 @@ def process_reply(self, beanstalkd_job_id, encoded_reply):
...
+                    beanstalkd_addr)
+                return None
+
+            if beanstalkd_jobs_ready >= self.max_jobs_per_beanstalkd:
+                self.logger.info(
+                    'Ignore beanstalkd %s: The queue is full (%d jobs)',
+                    beanstalkd_addr, beanstalkd_jobs_ready)
+                return None
+
+        beanstalkd.is_broken = False
```
loamhoof: That'd be cool to have a log in the case the beanstalkd was broken that says we're adding it back in the pool!

oio/xcute/orchestrator.py
loamhoof: A small error can happen anytime, it doesn't mean the beanstalkd is dead or dying. Maybe it's good enough for now, but I'd add a comment just saying that we could be more lenient and wait for a few errors in a row to happen before marking it as broken.

oio/xcute/orchestrator.py
```diff
@@ -170,15 +172,16 @@ def handle_running_job(self, job_info):
...
-                     beanstalkd_workers):
+    def dispatch_job(self, job_id, job_type, job_config, job_tasks):
         """
             Dispatch all of a job's tasks
         """
 
         self.logger.info('Start dispatching job (job_id=%s)', job_id)
 
         try:
+            beanstalkd_workers = self.get_beanstalkd_workers()
```
loamhoof: I forgot to mention it, but the iterator could be shared by all the jobs. Any pros/cons for one iterator per job? If we had one for all the jobs, the loadbalancing would probably be a bit more efficient

oio/xcute/orchestrator.py
AymericDu: Most jobs indirectly use the conscience (to know the address of service, to distribute the data on rawx, ...). But actually, we can consider these errors will be listed in the job information. In addition, there may be this problem of communication with conscience only for the orchestrator.

AymericDu: However, to continue sending tasks to a beanstalkd, we need to be sure that it still exists and has a positive score. If we don't do this verification, we could send tasks to a beanstalkd which should no longer receive them.

AymericDu: I suggest you add a comment to improve this point later.

oio/xcute/orchestrator.py
AymericDu: I like this idea and I had already thought about it. But the way I implemented it, I couldn't make it work with 2 jobs using this same iterator.
```
ValueError: generator already executing
```

PR #1931, opened 2019-11-21
xcute: Prefix request ID with job ID

AymericDu:
##### SUMMARY

Prefix request ID with job ID

##### ISSUE TYPE

 - Feature Pull Request

##### COMPONENT NAME

- xcute

##### SDS VERSION

```
openio 5.2.2
```

### DIFF
--- a/oio/xcute/common/worker.py
+++ b/oio/xcute/common/worker.py
@@ -15,10 +15,11 @@
 
 from collections import Counter
 
+from oio.common.constants import STRLEN_REQID
 from oio.common.green import ratelimit, sleep
 from oio.common.json import json
 from oio.common.logger import get_logger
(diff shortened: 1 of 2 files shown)


oio/xcute/common/worker.py
```diff
@@ -57,8 +58,10 @@ def process_beanstalkd_job(self, beanstalkd_job):
             tasks_run_time = ratelimit(
                     tasks_run_time, tasks_per_second)
 
+            reqid = job_id + request_id('-')
```
loamhoof: Shouldn't we simply use `request_id(job_id + '-')` instead?

AymericDu: The function returns 32 bytes and `job_id` already measures 32 bytes.

PR #1938, opened 2019-11-27
xcute: Keep tasks order to update tasks sent

AymericDu:
##### SUMMARY

Keep tasks order to update tasks sent, because the backend must have the tasks in order to know the last task sent

##### ISSUE TYPE

 - Bugfix Pull Request

##### COMPONENT NAME

- `xcute`

##### SDS VERSION

```
openio 5.2.2
```

### DIFF
--- a/oio/xcute/orchestrator.py
+++ b/oio/xcute/orchestrator.py
@@ -185,7 +185,9 @@ class XcuteOrchestrator(object):
             tasks_run_time = 0
             batch_per_second = tasks_per_second / float(
                 tasks_batch_size)
-            tasks = dict()
+            # The backend must have the tasks in order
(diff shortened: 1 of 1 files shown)


oio/xcute/orchestrator.py
```diff
@@ -186,11 +186,13 @@ def dispatch_job(self, job_id, job_type, job_config, job_tasks,
             batch_per_second = tasks_per_second / float(
                 tasks_batch_size)
             tasks = dict()
+            task_ids = list()
```
loamhoof: Use an `OrderedDict` instead: https://docs.python.org/2/library/collections.html#collections.OrderedDict

PR #1939, opened 2019-11-27
xcute:  Update processed tasks in workers

AymericDu:
##### SUMMARY

Update processed tasks in workers

Allow to:
- check if the job still exists
- check if the tasks are already processed (may occur during a beanstalkd crash)
- no longer use the reply tube (decreases the overhead)

##### ISSUE TYPE

 - Bugfix Pull Request

##### COMPONENT NAME

- xcute

##### SDS VERSION

```
openio 5.2.2
```


jkasarherou:
This PR is not a bugfix, it changes the whole orchestrator-worker design for imaginary reasons.

PR #1942, opened 2019-11-28
xcute: Check if tasks are already processed

AymericDu:
##### SUMMARY

Check if tasks are already processed

##### ISSUE TYPE

 - Bugfix Pull Request

##### COMPONENT NAME

- `xcute`

##### SDS VERSION

```
openio 5.2.2
```

### DIFF
--- a/oio/xcute/common/backend.py
+++ b/oio/xcute/common/backend.py
@@ -64,13 +64,13 @@ class XcuteBackend(RedisConnection):
         'job_finished': (
             Forbidden,
             'The job is finished'),
-        'job_already_waiting':  (
+        'job_already_waiting': (
             Forbidden,
             'The job is already waiting'),
-        'job_already_running':  (
+        'job_already_running': (
             Forbidden,
             'The job is already running'),
-        'job_running':  (
+        'job_running': (
             Forbidden,
             'The job running')
         }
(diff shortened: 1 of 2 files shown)


oio/xcute/orchestrator.py
```diff
@@ -212,9 +212,12 @@ def dispatch_job(self, job_id, job_type, job_config, job_tasks,
...
                 self.logger.info('All tasks sent (job_id=%s)' % job_id)
 
-                sent = self.dispatch_tasks(
-                    beanstalkd_workers,
-                    job_id, job_type, job_config, tasks)
+                if tasks:
+                    sent = self.dispatch_tasks(
+                        beanstalkd_workers,
+                        job_id, job_type, job_config, tasks)
+                else:
```
loamhoof: no need for the `else`, just set `sent = True` before the `if`

oio/xcute/common/backend.py
```diff
@@ -278,16 +281,19 @@ class XcuteBackend(RedisConnection):
...
-            'tasks.sent', tasks_sent_length);
-        for _, task_id in ipairs(tasks_sent) do
-            redis.call('SADD', 'xcute:tasks:running:' .. job_id, task_id);
-        end;
+        local nb_tasks_sent = 0;
         if tasks_sent_length > 0 then
+            nb_tasks_sent = redis.call(
+                'SADD', 'xcute:tasks:running:' .. job_id, unpack(tasks_sent));
+            if nb_tasks_sent ~= tasks_sent_length then
+                return redis.error_reply('tasks_already_sent');
```
loamhoof: isn't it broken because we don't update the number of tasks sent by `nb_tasks_sent`?

PR #1945, opened 2019-11-28
xcute: Improve error messages

AymericDu:
##### SUMMARY

- Convert all redis errors to excetions
- Catch all backend errors in the server

##### ISSUE TYPE

 - Bugfix Pull Request

##### COMPONENT NAME

- `xcute`

##### SDS VERSION

```
openio 5.2.2
```

### DIFF
--- a/oio/xcute/common/backend.py
+++ b/oio/xcute/common/backend.py
@@ -558,7 +558,7 @@ class XcuteBackend(RedisConnection):
         job_info = self.script_run_next(
             keys=[self._get_timestamp(), orchestrator_id],
             client=self.conn)
-        if job_info is None:
+        if not job_info:
             return None
 
         job_info = self._unmarshal_job_info(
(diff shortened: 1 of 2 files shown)


oio/xcute/common/backend.py
```diff
@@ -558,7 +558,7 @@ def run_next(self, orchestrator_id):
         job_info = self.script_run_next(
             keys=[self._get_timestamp(), orchestrator_id],
             client=self.conn)
-        if job_info is None:
+        if not job_info:
```
loamhoof: is it covering another case?

AymericDu: No, it isn't covering another case.
To avoid reproducing this error (https://github.com/open-io/oio-sds/pull/1945/files#diff-4a7c22a19c01798de356ba370cc8586bL632), I preferred to check `job_info` always in the same way.

PR #1946, opened 2019-12-02
Implement compression in rawx services

jfsmig:
##### SUMMARY
Update the golang RAWX to implement compression.

##### ISSUE TYPE
feature

##### COMPONENT NAME
rawx, tests


rawx/handler_chunk.go
```diff
@@ -275,24 +287,34 @@ func (rr *rawxRequest) checkChunk() {
...
-			LogError("Fail to compute md5sum: %s", err)
-			rr.replyError(err)
+			rr.replyCode(http.StatusPreconditionFailed)
 			return
 		}
-		if !strings.EqualFold(checksum, expected_hash) {
-			LogError("Md5sum doesn't match: computed:%s, expected:%s",
-				checksum, expected_hash)
-			rr.replyCode(http.StatusPreconditionFailed)
+		defer filter.Close()
```
AymericDu: When there is no compression, the filter is `nil`.
```suggestion
		if filter != nil {
			defer filter.Close()
		}
```

oio/blob/auditor.py
```diff
@@ -259,8 +261,14 @@ def _sleep(self):
 
 
 class ChunkReader(object):
-    def __init__(self, fp, size, md5_checksum):
+    def __init__(self, fp, size, md5_checksum, compression=None):
         self.fp = fp
+        if compression:
+            if compression == 'zlib':
```
murlock: We should support `deflate` and `lzw`

oio/common/constants.py
```diff
@@ -122,6 +123,7 @@
         'content_id': True,
         'content_path': True,
         'content_version': True,
+        'compression': True,
```
murlock: Please add an empty line before or move before l.121

PR #1948, opened 2019-12-03
client: allow https for proxy url

martinetd:
Do not assume the proxy url is plain text http

##### SUMMARY
We want to use https from the swift proxy to the oio proxy, but using an `https://` url as `sds_proxy_url` in `proxy-server.conf` does not work. This appears to fix the issue.

##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT NAME
oio client common

##### SDS VERSION
whatever it is master is? tested on 5.2.2

##### ADDITIONAL INFORMATION
Reproducer steps:
 - setup an https frontend to oioproxy (whatever reverse proxy that supports ssl terminaison) and configure `sds_proxy_url` to `https://<ip>:<port>` in `oioswift`'s `proxy-server.conf`
 - try to run a request, fails with internal error `No address found`, probably because the https is not stripped properly from the netloc and it cannot resolve it.
 - try again with patch, it should work.

Note: there is no SSL certificate validation, I would have assumed that change to fail a bit further.  
I would also like to validate the server certificate matches the one given in the url, any hint on where to change that would be appreciated before I start looking at the code.

### DIFF
--- a/oio/common/client.py
+++ b/oio/common/client.py
@@ -60,12 +60,15 @@ class ProxyClient(HttpApi):
             endpoint = self.conf.get('proxyd_url', None)
 
         ep_parts = list()
+        self.proxy_scheme = 'http'
         if endpoint:
-            self.proxy_netloc = endpoint.lstrip("http://")
+            if endpoint.startswith('https://'):
+                self.proxy_scheme = 'https'
+            self.proxy_netloc = endpoint.lstrip('%s://' % self.proxy_scheme)
         else:
             ns_conf = load_namespace_conf(self.ns)
             self.proxy_netloc = ns_conf.get('proxy')
-        ep_parts.append("http:/")
+        ep_parts.append('%s:/' % self.proxy_scheme)
         ep_parts.append(self.proxy_netloc)
 
         ep_parts.append("v3.0")
--- a/oio/container/client.py
+++ b/oio/container/client.py
@@ -113,7 +113,8 @@ class ContainerClient(ProxyClient):
         Build URIs for request that don't use the same prefix as the one
         set in this class' constructor.
         """
-        uri = 'http://%s/v3.0/%s/%s' % (self.proxy_netloc, self.ns, target)
+        uri = '%s://%s/v3.0/%s/%s' % (self.proxy_scheme, self.proxy_netloc,
+                                      self.ns, target)
         return uri
 
     def _make_params(self, account=None, reference=None, path=None, cid=None,


martinetd:
> travis failure

seems to be unrelated, master is failing on the same test ; service not ready yet looks like a scheduling problem...

> Note: there is no SSL certificate validation, I would have assumed that change to fail a bit further.
> I would also like to validate the server certificate matches the one given in the url, any hint on where to change that would be appreciated before I start looking at the code.

This stems from urllib3 default. until 1.25 (el7.7 has 1.21) certificate validation was disabled by default.
One would need to add a couple of arguments to the pool manager (cert_reqs='CERT_REQUIRED' *and* a path to valid ca_certs... You'd think that since the python ssl module has a default ca_certs path it could pick it on its own :/
This is a more intrusive change than just allowing https so while I got something working I'll need to figure how to add this cleanly to some config and will submit a different PR when ready.

martinetd:
(hmpf github doesn't let one request changes on one's own PR, bad system bad)

I was only testing this with metadata operations (e.g. ls), now I've finished splitting services a bit more cleanly it turns out some data is still sent as plain text http on the https port with just this change, I must have missed something... Will repush/update once I find out more...

martinetd:
>  I was only testing this with metadata operations (e.g. ls), now I've finished splitting services a bit more cleanly it turns out some data is still sent as plain text http on the https port with just this change, I must have missed something... Will repush/update once I find out more...

now works for basic tests at least; there are still hardcoded `http://`s associated with `proxy_netloc` in `oio/directory/admin.py` that might need similar fix but I don't know how to test this bit (yet)
`oio/cli/common/clientmanager.py` is just setting default values so is probably fine as is.

The more I look at this and the less I understand why `proxy_netloc` does not embed the scheme directly in the variable, as it always seems to be used concatenated with `http://` stuck in front -- what was the intended difference between the endpoint and that variable?

fvennetier:
Hello! Thanks for your contribution.

> The more I look at this and the less I understand why `proxy_netloc` does not embed the scheme directly in the variable, as it always seems to be used concatenated with `http://` stuck in front -- what was the intended difference between the endpoint and that variable?

Yes. Here "netloc" has the same meaning as in the result of `urllib.parse.urlparse()`. Endpoint is scheme+netloc+path_prefix.

And yes, our test suite on Travis CI in kind of flaky. We see less random failures when we run the tests before 14:30 UTC (because the platform is less loaded).

martinetd:
> Could you rebase on master one last time? I have fixed the test suite.

Done! Let's see how it goes, thanks.

martinetd:
FWIW this does not cover the connections between the swift gateway and the rawx, we will need to upgrade these to https as well somehow to meet our requirements.

I'm not sure if you want that in the same PR or not, but I would appreciate some help on this last point as my development time is fairly limited; I will open an issue to track that.

martinetd:
```
_StringException: Traceback (most recent call last):
  File "/home/travis/build/open-io/oio-sds/tests/functional/content/test_content_perfectible.py", line 143, in test_upload_warn_dist
    event = self._wait_for_event(timeout=REASONABLE_EVENT_DELAY*2)
  File "/home/travis/build/open-io/oio-sds/tests/functional/content/test_content_perfectible.py", line 96, in _wait_for_event
    self.fail("No event received in the last %s seconds" % timeout)
  File "/home/travis/build/open-io/oio-sds/.tox/func/local/lib/python2.7/site-packages/unittest2/case.py", line 690, in fail
    raise self.failureException(msg)

AssertionError: No event received in the last 6.0 seconds
```

That looks like it could be a load issue, I'll try to resubmit both PRs somehow tomorrow morning

martinetd:
> That looks like it could be a load issue, I'll try to resubmit both PRs somehow tomorrow morning

No luck, exact same error. I'm not quite sure how it could be related, the patch should have no effect on existing config (same variables defined with same content); I'll leave that up to you.

Thanks.

martinetd:
> No luck, exact same error. I'm not quite sure how it could be related, the patch should have no effect on existing config (same variables defined with same content); I'll leave that up to you.

Obviously just had to post to see updated status (sorry for the spam); I cannot seem to access the codecov report using a dummy account and do not want to give access to all the permissions it asks for for this account, I would appreciate a retranscription or feedback. Sorry for the trouble.

fvennetier:
> I cannot seem to access the codecov report using a dummy account and do not want to give access to all the permissions it asks for for this account, I would appreciate a retranscription or feedback.

Codecov is supposed to post a comment here once all the tests are green. But sometimes it does not. Today all I can get is a 500 error. I will merge this anyway.

PR #1953, opened 2019-12-11
xcute: Retry later when the worker can't reply

AymericDu:
##### SUMMARY

Retry later when the worker can't reply

##### ISSUE TYPE

 - Bugfix Pull Request

##### COMPONENT NAME

- `xcute`

##### SDS VERSION

```
openio 5.2.3.dev1
```

### DIFF
--- a/oio/event/filters/xcute.py
+++ b/oio/event/filters/xcute.py
@@ -13,19 +13,60 @@
 # You should have received a copy of the GNU Affero General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
-from oio.event.evob import Event
+from oio.event.evob import Event, EventTypes
+from oio.event.beanstalk import BeanstalkError
 from oio.event.filters.base import Filter
+from oio.common.easy_value import int_value
+from oio.common.json import json
 from oio.xcute.common.worker import XcuteWorker
 
 
 class XcuteFilter(Filter):
 
+    DEFAULT_RETRY_DELAY_TO_REPLY = 60
+
     def init(self):
+        self.retry_delay_to_reply = int_value(
+            self.conf.get('retry_delay_to_reply'),
+            self.DEFAULT_RETRY_DELAY_TO_REPLY)
+
         self.worker = XcuteWorker(self.conf, logger=self.logger)
 
     def process(self, env, cb):
         event = Event(env)
-        self.worker.process_beanstalkd_job(event.data)
+
+        if event.data.get('processed'):
+            job_id = event.data['job_id']
+            task_ids = event.data['task_ids']
+            task_results = event.data['task_results']
+            task_errors = event.data['task_errors']
+            beanstalkd_reply_info = event.data['beanstalkd_reply']
+        else:
+            job_id, task_ids, task_results, task_errors, \
+                beanstalkd_reply_info = self.worker.process(event.data)
+
+        try:
+            self.worker.reply(
+                job_id, task_ids, task_results, task_errors,
+                beanstalkd_reply_info)
+        except BeanstalkError as exc:
+            self.logger.warn(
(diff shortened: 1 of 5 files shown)


oio/xcute/common/worker.py
```diff
@@ -80,16 +80,13 @@ def _reply(self, reply_addr, reply_tube,
...
+            beanstalkd_reply.watch(self.beanstalkd_workers_tube)
 
-        beanstalkd_sender = self.beanstalkd_senders[(reply_addr, reply_tube)]
+            self.beanstalkd_replies[beanstalkd_reply_info] = beanstalkd_reply
 
-        while not beanstalkd_sender.send_job(reply_payload):
-            sleep(1)
-
-        beanstalkd_sender.job_done()
+        beanstalkd_reply.put(reply_payload)
```
loamhoof: What happens when the `put` fails? Does it raise an exception? If so, what does the event-agent do?

oio/event/filters/xcute.py
```diff
@@ -13,19 +13,64 @@
...
+            job_id = event.data['job_id']
+            task_ids = event.data['task_ids']
+            task_results = event.data['task_results']
+            task_errors = event.data['task_errors']
+            beanstalkd_reply_info = event.data['beanstalkd_reply']
+        else:
+            job_id, task_ids, task_results, task_errors, \
+                beanstalkd_reply_info = self.worker.process(event.data)
+
+        if beanstalkd_reply_info:
```
loamhoof: In what case wouldn't we have one?

oio/event/filters/xcute.py
```diff
@@ -13,19 +13,64 @@
...
+        beanstalkd_reply_info = None
+        if event.data.get('processed'):
+            job_id = event.data['job_id']
+            task_ids = event.data['task_ids']
+            task_results = event.data['task_results']
+            task_errors = event.data['task_errors']
+            beanstalkd_reply_info = event.data['beanstalkd_reply']
+        else:
+            job_id, task_ids, task_results, task_errors, \
+                beanstalkd_reply_info = self.worker.process(event.data)
```
loamhoof: `process` shouldn't have to return `beanstalkd_reply_info` anymore. This function could do it instead.

oio/xcute/common/worker.py
```diff
@@ -68,25 +65,29 @@ def process_beanstalkd_job(self, beanstalkd_job):
...
-               job_id, task_ids, task_results, task_errors):
-        reply_payload = json.dumps({
-            'job_id': job_id,
-            'task_ids': task_ids,
-            'task_results': task_results,
-            'task_errors': task_errors
-        })
+    def reply(self, job_id, task_ids, task_results,
+              task_errors, beanstalkd_reply_info):
+        if not beanstalkd_reply_info:
```
loamhoof: You already made this check before?

oio/xcute/common/worker.py
AymericDu: If this `put` fails, we catch exception here: https://github.com/open-io/oio-sds/pull/1953/files#diff-8cd3ec0f8415e064fe679869ed2e03a8R52

AymericDu: Otherwise, if `XcuteFilter.process` raises an exception, the event will be buried (this is the behavior of the event-agent).

oio/event/filters/xcute.py
AymericDu: I prefer to isolate the response (and these errors) to manage the following 2 cases in the same place:
- the first response attempt (just after processing the tasks)
- the following attempts (as long as it is not possible to reply)

oio/xcute/common/worker.py
loamhoof: :+1:

PR #1955, opened 2019-12-13
proxy: Allow to force the request on master service

AymericDu:
##### SUMMARY

Allow to force the request on master service.

It was already possible to do this by adding 'proxy.force.master'
in the proxy configuration. But this was applied to all requests.

If we wanted to force the master only on a specific request,
we can add the 'x-oio-force-master' header in the request.
But this header was only used to list the objects in a container.

Now, it is possible to specifically force the master on all meta2 requests.

##### ISSUE TYPE

 - Bugfix Pull Request

##### COMPONENT NAME

- `proxy`
- `meta2`

##### SDS VERSION

```
openio 5.2.4.dev1
```

### DIFF
--- a/meta2v2/meta2_backend.c
+++ b/meta2v2/meta2_backend.c
@@ -84,7 +84,7 @@ static void m2b_add_modified_container(struct meta2_backend_s *m2b,
 static enum m2v2_open_type_e
 _mode_masterslave(guint32 flags)
 {
-	if (flags & M2V2_FLAG_MASTER)
+	if ((flags & M2V2_FLAG_MASTER) || oio_ext_has_force_master())
 		return M2V2_OPEN_MASTERONLY;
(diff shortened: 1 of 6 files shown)


meta2v2/meta2_filters_action_container.c
```diff
@@ -309,11 +309,11 @@ _load_list_params(struct list_params_s *lp, struct gridd_filter_ctx_s *ctx,
 	const char *fstr = meta2_filter_ctx_get_param(ctx, NAME_MSGKEY_FLAGS);
 	if (NULL != fstr) {
 		guint32 flags = atoi(fstr);
+		lp->flags = flags;
```
fvennetier: Instead of merging all booleans in an obscure "flags" field, I vote for generalizing the thread-local structure usage. This means calling `meta2_filter_extract_force_master` in all meta2 request handlers doing read-only operations.

PR #1956, opened 2019-12-13
xcute: Handle backend errors

AymericDu:
##### SUMMARY

Handle backend errors

##### ISSUE TYPE

 - Bugfix Pull Request

##### COMPONENT NAME

- `xcute`

##### SDS VERSION

```
openio 5.2.4.dev114
```

### DIFF
--- a/oio/xcute/orchestrator.py
+++ b/oio/xcute/orchestrator.py
@@ -16,6 +16,8 @@
 import math
 import random
 from collections import OrderedDict
+from redis import ConnectionError as RedisConnectionError, \
+    TimeoutError as RedisTimeoutError
 
 from oio.common.easy_value import int_value
 from oio.common.exceptions import OioTimeout
@@ -73,6 +75,28 @@ class XcuteOrchestrator(object):
         self.beanstalkd_workers = dict()
         self.threads = dict()
 
+    def handle_backend_errors(self, func, *args, **kwargs):
+        while True:
+            try:
+                return func(*args, **kwargs), None
+            except (RedisConnectionError, RedisTimeoutError) as exc:
+                self.logger.warn(
+                    'Fail to communicate with redis: %s', exc)
+                if not self.running:
+                    return None, exc
+                sleep(1)
+
+    def safe_run_forever(self):
+        try:
+            self.run_forever()
+        except Exception as exc:
+            self.logger.exception('Fail to run forever: %s', exc)
+            self.exit_gracefully()
+
+        for thread_ in self.threads.values():
+            thread_.join()
+        self.logger.info('Exited running thread')
+
     def run_forever(self):
         """
             Take jobs from the queue and spawn threads to dispatch them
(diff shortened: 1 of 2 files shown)


oio/xcute/orchestrator.py
```diff
@@ -80,22 +93,23 @@ def run_forever(self):
...
-
         # start processing replies
         listen_thread = threading.Thread(target=self.listen)
         listen_thread.start()
-
         self.threads[listen_thread.ident] = listen_thread
 
+        # restart running jobs
+        orchestrator_jobs = None
+        if self.running:
```
loamhoof: Early return
There is no point putting everything inside ifs

oio/xcute/orchestrator.py
```diff
@@ -59,6 +61,17 @@ def __init__(self, conf, logger=None):
...
 
+    def handle_backend_errors(self, func, *args, **kwargs):
+        while True:
+            try:
+                return func(*args, **kwargs)
+            except (RedisConnectionError, RedisTimeoutError) as exc:
+                self.logger.error(
+                    'Fail to communicate with redis: %s', exc)
+                if not self.running:
+                    return
```
loamhoof: You're changing the return value of `func`
This is risky

oio/xcute/orchestrator.py
```diff
@@ -203,9 +218,11 @@ def dispatch_job(self, job_id, job_type, job_config, job_tasks,
                     beanstalkd_workers,
                     job_id, job_type, job_config, tasks)
                 if sent:
-                    job_status = self.backend.update_tasks_sent(
-                        job_id, tasks.keys())
+                    job_status = self.handle_backend_errors(
+                        self.backend.update_tasks_sent, job_id, tasks.keys())
                     tasks.clear()
+                    if job_status is None:  # self.running == False
```
loamhoof: Consequence of changing the return value of the function.

oio/xcute/orchestrator.py
```diff
@@ -126,7 +141,7 @@ def orchestrate_loop(self):
             except Exception:
                 self.logger.exception(
                     'Failed to instantiate job %s', job_id)
-                self.backend.fail(job_id)
+                self.handle_backend_errors(self.backend.fail(job_id))
```
loamhoof: That doesn't work.

oio/xcute/orchestrator.py
```diff
@@ -117,7 +131,8 @@ def orchestrate_loop(self):
         """
 
         new_jobs = iter(
-            lambda: self.backend.run_next(self.orchestrator_id), None)
+            lambda: self.handle_backend_errors(
+                self.backend.run_next(self.orchestrator_id), None))
```
loamhoof: That doesn't work.

oio/xcute/orchestrator.py
```diff
@@ -216,21 +233,22 @@ def dispatch_job(self, job_id, job_type, job_config, job_tasks,
...
-                    return
-
-            self.backend.free(job_id)
-        except Exception:
-            self.logger.exception('Failed generating task list (job_id=%s)',
-                                  job_id)
+                    job_status = self.handle_backend_errors(
+                        self.backend.update_tasks_sent, job_id, tasks.keys(),
+                        all_tasks_sent=True)
+                    if job_status is not None:  # self.running == True
```
loamhoof: Same

loamhoof:
Overall a pretty heavy change for the code. Can't we just catch the exception at a higher level and put the retry at a lower one?

oio/xcute/orchestrator.py
AymericDu: If we return here, we don't wait for `beanstalkd_thread` and `listen_thread` to stop. To exit properly, we have to wait for all started threads to stop.

loamhoof: @AymericDu You can put this code inside a function and wait for the threads to stop outside the function?

oio/xcute/orchestrator.py
```diff
@@ -116,8 +136,15 @@ def orchestrate_loop(self):
...
-            lambda: self.backend.run_next(self.orchestrator_id), None)
+        def run_next_job():
+            next_job, exc = self.handle_backend_errors(
+                self.backend.run_next, self.orchestrator_id)
+            if exc is None:
+                return next_job
+            self.logger.warn('Unable to run next job: %s', exc)
+            return None
+
+        new_jobs = iter(lambda: run_next_job(), None)
```
loamhoof: If the code has become this complicated, I don't think there's any point in using `iter(iterable, None)` anymore, is it?

PR #1972, opened 2020-01-16
CLI: check peer count before actually rebalancing meta1

fvennetier:
##### SUMMARY
During a rebalance operation, check each reference prefix is managed by the right number of meta1 services.

Also implement a new 'openio directory check' command, checking each reference prefix is managed by the right number of meta1 services. This complements 'openio-admin directory check' (which does not check this number).

Jira: OS-470

##### ISSUE TYPE
 - Feature Pull Request
 - Bugfix Pull Request

##### COMPONENT NAME
- CLI

##### SDS VERSION
```
openio 5.3.0.dev3
```

### DIFF
--- a/oio/cli/directory/directory.py
+++ b/oio/cli/directory/directory.py
@@ -1,4 +1,4 @@
-# Copyright (C) 2015-2019 OpenIO SAS, as part of OpenIO SDS
+# Copyright (C) 2015-2020 OpenIO SAS, as part of OpenIO SDS
 #
 # This program is free software: you can redistribute it and/or modify
 # it under the terms of the GNU Affero General Public License as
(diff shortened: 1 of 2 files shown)


oio/cli/directory/directory.py
```diff
@@ -196,9 +218,12 @@ def take_action(self, parsed_args):
...
         moved = mapping.rebalance()
-        self._apply(mapping, moved=moved,
-                    read_timeout=parsed_args.meta0_timeout)
-        self.log.info("Moved %s", moved)
+        if mapping.check_replicas():
+            self._apply(mapping, moved=moved,
+                        read_timeout=parsed_args.meta0_timeout)
+            self.log.info("Moved %s", moved)
+        else:
+            self.log.warn("Nothing done due to errors")
```
AymericDu: We should also add `self.success = False`.

PR #1975, opened 2020-01-20
xcute: locks feature

loamhoof:
##### SUMMARY
Add a lock system to xcute to avoid collisions between jobs.
Each job type defines its lock based on its parameters. The lock is taken when the job is created, and released when it finishes or fails.

##### ISSUE TYPE
 - Feature Pull Request

##### COMPONENT NAME
xcute

##### ADDITIONAL INFORMATION
Results of the CLI commands:
```
(venv) openio@f12651ea56b2:/tmp/src/oio-sds$ openio-admin xcute lock list
+------+----------------------------------+
| Lock | Job ID                           |
+------+----------------------------------+
| a    | 20200120074545861743-802a3d50197 |
| aa   | 20200120074838579195-f4d429568d3 |
| abb  | 20200120074937560447-51af50063c5 |
| b    | 20200120074547476397-d5c7e58d0af |
+------+----------------------------------+
(venv) openio@f12651ea56b2:/tmp/src/oio-sds$ openio-admin xcute lock show abb
+--------+----------------------------------+
| Field  | Value                            |
+--------+----------------------------------+
| lock   | abb                              |
| job_id | 20200120074937560447-51af50063c5 |
+--------+----------------------------------+
```

Error message when trying to create a job which lock is already taken:

```
jcurl -s localhost:8002/v1.0/xcute/job/create\?type=tester -d '{"tasks_per_second": 1, "params": {"end": 100, "lock
": "aab"}}'
A job with the same lock (aab) is already in progress
```

### DIFF
--- a/oio/xcute/common/backend.py
+++ b/oio/xcute/common/backend.py
@@ -49,6 +49,9 @@ class XcuteBackend(RedisConnection):
         'job_exists': (
             Forbidden,
             'The job already exists'),
+        'lock_exists': (
+            Forbidden,
+            'A job with the same lock is already in progress'),
         'no_job': (
             NotFound,
             'The job does\'nt exist'),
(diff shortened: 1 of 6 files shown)


oio/xcute/common/backend.py
```diff
@@ -641,6 +669,21 @@ def get_job_info(self, job_id):
...
         return client.hgetall(self.key_job_info % job_id)
 
+    @handle_redis_exceptions
+    def list_locks(self):
+        locks = self.conn.hgetall(self.key_locks)
+
+        return [
+            dict(lock=lock[0], job_id=lock[1])
+            for lock in sorted(locks.items())
+        ]
```
AymericDu: Why return a list of dictionaries `[{"lock": LOCK, "job_id": JOB_ID}, ...]` instead of the dictionary `{LOCK: JOB_ID, ...}`?
Also, if the client wants to sort the locks, it can do it itself. This will allow a faster response and free up the server.

PR #1977, opened 2020-01-24
Pass options for `ca_certs` and `cert_reqs`

martinetd:
##### SUMMARY
urllib3 request pool does not validate server certificate by default for https connections until version 1.25 (at which point we would want to specify ca_certs anyway)

This just allows infrastructure to pass the option through and add `ca_certs` and `cert_reqs` to parameter whitelist.

##### ISSUE TYPE
 - Feature Pull Request

##### COMPONENT NAME
common?

##### ADDITIONAL INFORMATION
This builds on top of #1948, so should only be merged after that.
There is no way to do https without anyway so it won't be missed :)

(also note I didn't test the api/base part as gateway swift only uses ClientManager, but figured I might as well try that..)

### DIFF
--- a/oio/api/base.py
+++ b/oio/api/base.py
@@ -20,19 +20,12 @@ from oio.common.easy_value import true_value
 from oio.common.json import json as jsonlib
 
 from oio.common.http_urllib3 import urllib3, get_pool_manager, \
-    oio_exception_from_httperror
+    oio_exception_from_httperror, URLLIB3_REQUESTS_KWARGS
 from oio.common import exceptions
 from oio.common.utils import deadline_to_timeout, monotonic_time
 from oio.common.constants import ADMIN_HEADER, \
     TIMEOUT_HEADER, PERFDATA_HEADER, FORCEMASTER_HEADER, \
     CONNECTION_TIMEOUT, READ_TIMEOUT, REQID_HEADER, STRLEN_REQID
-
-_POOL_MANAGER_OPTIONS_KEYS = ["pool_connections", "pool_maxsize",
-                              "max_retries"]
-
-URLLIB3_REQUESTS_KWARGS = ('fields', 'headers', 'body', 'retries', 'redirect',
-                           'assert_same_host', 'timeout', 'pool_timeout',
-                           'release_conn', 'chunked')
 
 
 class HttpApi(object):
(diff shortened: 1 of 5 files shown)


oio/api/base.py
```diff
@@ -20,20 +20,14 @@
 from oio.common.json import json as jsonlib
 
 from oio.common.http_urllib3 import urllib3, get_pool_manager, \
-    oio_exception_from_httperror
+    oio_exception_from_httperror, URLLIB3_POOLMANAGER_KWARGS,
```
fvennetier:
```suggestion
    oio_exception_from_httperror, URLLIB3_POOLMANAGER_KWARGS, \
```

martinetd: Good catch, I didn't test this part as I found out later the swift gateway uses only clientmanager

oio/api/base.py
```diff
@@ -63,7 +57,7 @@ def __init__(self, endpoint=None, pool_manager=None,
         if not pool_manager:
             pool_manager_conf = {k: int(v)
                                  for k, v in iteritems(kwargs)
-                                 if k in _POOL_MANAGER_OPTIONS_KEYS}
+                                 if k in URLLIB3_POOLMANAGER_KWARGS}
```
martinetd: Ah, it took me a while to understand why tests failed here with that:
```
ERROR: tests.functional.api.test_objectstorage_performance.TestObjectStorageApiPerformance.test_object_create_32_md5_checksum
----------------------------------------------------------------------
_StringException: Traceback (most recent call last):
  File "/home/travis/build/open-io/oio-sds/tests/functional/api/test_objectstorage_performance.py", line 28, in setUp
    source_address=('127.0.0.8', 0))
  File "oio/api/object_storage.py", line 110, in __init__
    self.directory = DirectoryClient(conf, logger=self.logger, **kwargs)
  File "oio/directory/client.py", line 30, in __init__
    **kwargs)
  File "oio/common/client.py", line 83, in __init__
    endpoint='/'.join(ep_parts), service_type='proxy', **kwargs)
  File "oio/api/base.py", line 59, in __init__
    for k, v in iteritems(kwargs)
  File "oio/api/base.py", line 60, in <dictcomp>
    if k in URLLIB3_POOLMANAGER_KWARGS}
TypeError: int() argument must be a string or a number, not 'tuple'
```

The reason is the old list only allowed integer arguments and using this straight from URLLIB3_POOLMANAGER_KWARGS is more liberal.
Looking at git log, python3 compat might have required the `int()` cast? but 7966a1e63c2850f3e5c2faa494d2d0f9ff4adf26 does not say why limit to int args and not pass-through non-int args, and `get_pool_manager` now does the cast and filtering itself...

So I'll just drop the whole filtering back out and assume the function does it properly, and see how tests hold up.

PR #1978, opened 2020-01-24
rawx: fix behavior when chunk size attr is missing

fvennetier:
##### SUMMARY
When the chunk size extended attribute was missing, the rawx service used to fail while trying to parse an integer from an empty string (`Failed to load xattr: strconv.ParseInt: parsing "": invalid syntax`). Now, when the attribute is missing, it won't try the integer conversion, and simply return the expected code 412.

##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT NAME
- rawx

##### SDS VERSION
```
6.1.0.0a0-29-gf0b4a00d
```

### DIFF
--- a/rawx/chunk_info.go
+++ b/rawx/chunk_info.go
@@ -226,10 +226,11 @@ func (chunk *chunkInfo) loadAttr(inChunk fileReader, chunkID string,
 		}
 		*(hs.ptr) = value
 	}
-
-	chunk.size, err = strconv.ParseInt(chunk.ChunkSize, 10, 63)
-	if err != nil {
-		return err
+	if chunk.ChunkSize != "" {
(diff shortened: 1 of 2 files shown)


rawx/chunk_info.go
```diff
@@ -226,10 +226,11 @@ func (chunk *chunkInfo) loadAttr(inChunk fileReader, chunkID string,
 		}
 		*(hs.ptr) = value
 	}
-
-	chunk.size, err = strconv.ParseInt(chunk.ChunkSize, 10, 63)
-	if err != nil {
-		return err
+	if chunk.ChunkSize != "" {
```
murlock: A warning is raised in log but if attribute is missing, the GET of chunk will return 0 bytes without error.
IMO a return `fmt.Errorf` should be added in this case

PR #1981, opened 2020-01-29
Xcute: fix the orchestrator using all the beanstalkd

loamhoof:
##### SUMMARY
The xcute orchestrator is currently using all the beanstalkd available, instead of only using the ones with the `oio-xcute` tube. This can lead to tasks that will never be treated and hence never-ending jobs.

##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT NAME
xcute

### DIFF
--- a/oio/xcute/orchestrator.py
+++ b/oio/xcute/orchestrator.py
@@ -456,10 +456,14 @@ class XcuteOrchestrator(object):
 
             old_beanstalkd_workers_addr = set(self.beanstalkd_workers.keys())
             new_beanstalkd_workers_addr = set(beanstalkd_workers.keys())
+
             added_beanstalkds = new_beanstalkd_workers_addr \
                 - old_beanstalkd_workers_addr
             for beanstalkd_addr in added_beanstalkds:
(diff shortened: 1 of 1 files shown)


oio/xcute/orchestrator.py
```diff
@@ -510,8 +510,6 @@ def _check_beanstalkd_worker(self, beanstalkd_info):
         if not beanstalkd:
             beanstalkd = Beanstalk.from_url(beanstalkd_addr)
             beanstalkd.addr = beanstalkd_addr
-            beanstalkd.use(self.beanstalkd_workers_tube)
-            beanstalkd.watch(self.beanstalkd_workers_tube)
```
AymericDu: Indeed, the use of the tube requested too early. But these methods must still be called to specifically use this tube. We could do it after all the checks (https://github.com/open-io/oio-sds/pull/1981/files#diff-965c127dbd1971b4f3ad28a7884213eaL544).

loamhoof: My bad, for some reason, without doing `use`, a job is still dispatched without raising an error (that's weird). But why do we need the `watch`? We're not reading from the tube.

PR #2000, opened 2020-03-10
Enable or disable replication for each bucket

fvennetier:
##### SUMMARY
Jira: OB-477

##### ISSUE TYPE
 - Feature Pull Request

##### COMPONENT NAME
- account service
- event-agent

##### SDS VERSION
```
openio 5.3.0.0b1.dev31
```

### DIFF
--- a/oio/account/backend.py
+++ b/oio/account/backend.py
@@ -20,6 +20,7 @@ import redis
 import redis.sentinel
 import six
 from werkzeug.exceptions import NotFound, Conflict, BadRequest
+from oio.common.constants import BUCKET_PROP_REPLI_ENABLED
 from oio.common.timestamp import Timestamp
 from oio.common.easy_value import int_value, boolean_value, float_value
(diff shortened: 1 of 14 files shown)


oio/account/backend.py
```diff
@@ -314,6 +315,26 @@ class AccountBackend(RedisConnection):
         end;
         """
 
+    lua_get_extended_container_info = """
+        local ckey = KEYS[1]
+        local bucket_prefix = ARGV[1]
+        local bname = redis.call('HGET', ckey, 'bucket')
+
+        local replication_enabled = 'false'
```
AymericDu: Instead of adding an option (`check_replication_enabled`), why not set the default value to true?

fvennetier: If we enable it by default, any unconfigured bucket will be replicated. The user would have to explicitly disable it for each bucket. That's not what we want.

And also, by disabling the check, legacy platforms won't see any performance degradation.

PR #2004, opened 2020-03-20
Allow creating object with TLS rawx

murlock:
##### SUMMARY

This PR will allow Python API (and gateway) to read and write object with TLS HTTP connections.

This PR covers:
- rawx with dedicated IP:PORT with TLS
- proxy will retrieves information with all tags (it may be updated to fetch only 'static' tags)
- when a client want to use TLS to rawx, it must provide upgrade-to-tls header to inform proxy to provide `real_url` with link IF rawx exposes themselves a HTTPS

This PR does not cover certificate verification. 

Fixes OB-502

##### ISSUE TYPE
 - Feature Pull Request

##### COMPONENT NAME
rawx, proxy, Python API, cli

##### SDS VERSION
```
6.x
```


oio/common/client.py
```diff
@@ -95,6 +95,12 @@ def _direct_request(self, method, url, headers=None, request_attempts=None,
             headers["X-oio-action-mode"] = "autocreate"
             kwargs = kwargs.copy()
             kwargs.pop("autocreate")
+        if kwargs.get("tls"):
+            if not headers:
+                headers = dict()
+            kwargs = kwargs.copy()
+            headers["X-oio-upgrade-to-tls"] = kwargs.pop("tls")
```
murlock: I will use `HEADER_PREFIX` instead and fix l.95

oio/cli/cluster/cluster.py
```diff
@@ -89,32 +93,34 @@ def _list_services(self, parsed_args):
...
-                volume = tags.get('tag.vol', 'n/a')
-                service_id = tags.get('tag.service_id', 'n/a')
+                location = tags.pop('tag.loc', 'n/a')
+                slots = tags.pop('tag.slots', 'n/a')
+                volume = tags.pop('tag.vol', 'n/a')
+                service_id = tags.pop('tag.service_id', 'n/a')
                 addr = srv['addr']
-                locked = boolean_value(tags.get('tag.lock'), False)
-                up = tags.get('tag.up', 'n/a')
+                locked = boolean_value(tags.pop('tag.lock'), False)
```
fvennetier: We should also set a default value as second argument to pop, or it may raise KeyError.

oio/common/client.py
```diff
@@ -92,9 +93,15 @@ def _direct_request(self, method, url, headers=None, request_attempts=None,
...
             if not headers:
                 headers = dict()
-            headers["X-oio-action-mode"] = "autocreate"
+            headers[HEADER_PREFIX + "action-mode"] = "autocreate"
             kwargs = kwargs.copy()
             kwargs.pop("autocreate")
+        if kwargs.get("tls"):
+            if not headers:
+                headers = dict()
+            kwargs = kwargs.copy()
```
fvennetier: I'm not sure this copy is necessary.

rawx/main.go
```diff
@@ -228,25 +224,46 @@ func main() {
 		MaxHeaderBytes: opts.getInt("headers_buffer_size", 65536),
 	}
 
+	srvTLS := http.Server{
```
fvennetier: tlsSrv?

rawx/rawx.go
```diff
@@ -35,6 +35,7 @@ func setError(rep http.ResponseWriter, e error) {
 type rawxService struct {
 	ns           string
 	url          string
+	urlTLS       string
```
fvennetier: tlsUrl?

tools/server.pem
```diff
@@ -0,0 +1,49 @@
+-----BEGIN PRIVATE KEY-----
```
fvennetier: We should insert a comment telling this is a test file (and has been put here on purpose).

proxy/m2_actions.c
```diff
@@ -209,9 +209,20 @@ _real_url_from_chunk_id (const char *id)
 	oio_parse_chunk_url(id, &type, &netloc, NULL);
 	addr = _resolve_service_id(netloc);
 
-	/* generate real_url, only if item was found */
+	/* TODO(mbo): fix service/IP */
+	if (oio_ext_has_upgrade_to_tls()) {
+		gchar *tls = get_tls(addr);
```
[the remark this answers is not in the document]
murlock: Code is no more available in last commit, I'll update this if required

proxy/metacd_http.c
```diff
@@ -1263,6 +1208,7 @@ grid_main_configure (int argc, char **argv)
 	lb_world_rawx = oio_lb_local__create_world();
 	lb_rawx = oio_lb__create();
 
+
```
AymericDu: I think this blank line was added by mistake.

rawx/http.go
```diff
@@ -0,0 +1,53 @@
...
+				errs <- err
+			}
+		}()
+	}
+
+	select {
+	case err := <-errs:
+		log.Printf("Could not start serving service due to (error: %s)", err)
+		return err
+	}
```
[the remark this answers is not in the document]
murlock: :+1:

oio/common/client.py
murlock: This copy is unnecessary if the `autocreate` is set but still necessary for other case.
I've updated to avoid the unnecessary copy

fvennetier: The copy is not necessary because the kwargs dictionary is guaranteed to be new. When we pop elements out of it, the dictionary seen by the calling method will still see them.
https://stackoverflow.com/questions/45883655/is-it-always-safe-to-modify-the-kwargs-dictionary

core/lb.c
```diff
@@ -284,6 +303,7 @@ struct oio_lb_slot_s
  * - all the slots that gather services with the same characteristics
  * - generation of the services in the 'world' (incr by 1 at each reload)
  * - absolute maximum distance between services in the 'world'
+ * - callbacks when adding/removing item
```
fvennetier, line 306: I think you got rid of that.

PR #2010, opened 2020-03-26
rawx: ignore errors for few xattr

murlock:
##### SUMMARY

When a chunk is uploaded in other storage_policy than EC,
the metachunk-size and metachunk-hash is no sent by client.

When rawx read this chunk, it complains about missing xattr.

This PR will ignore only these xattr.

##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT NAME
rawx

##### SDS VERSION
```
master
```

### DIFF
--- a/rawx/chunk_info.go
+++ b/rawx/chunk_info.go
@@ -163,6 +163,7 @@ func (chunk *chunkInfo) loadAttr(inChunk fileReader, chunkID string,
 		}
 	}
 
+	/* keep AttrNameContentStgPol above AttrNameMetachunkChecksum and AttrNameMetachunkSize */
 	var detailedAttrs = []detailedAttr{
 		{AttrNameContentChunkMethod, &chunk.ContentChunkMethod},
 		{AttrNameContentStgPol, &chunk.ContentStgPol},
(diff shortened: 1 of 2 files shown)


rawx/chunk_info.go
```diff
@@ -216,16 +217,20 @@ func (chunk *chunkInfo) loadAttr(inChunk fileReader, chunkID string,
 	chunk.ChunkID = chunkID
 
 	for _, hs := range detailedAttrs {
-		value, err := getAttr(hs.key)
+		*(hs.ptr), err = getAttr(hs.key)
```
fvennetier: In case of error, `*(hs.ptr)` will be initialized with an empty string instead of staying uninitialized. Is it wanted?

murlock: Please not that getAttr always return empty string in case of error.
This behavior was already the case before this PR

fvennetier: > Please not that getAttr always return empty string in case of error.
> This behavior was already the case before this PR

Yes, but what about `*(hs.ptr)`?

murlock: `ptr` reference chunk fields that was previously initialized empty by ServeHTTP method (rawx.go l.118 ) when rawRequest is instancied

PR #2016, opened 2020-04-09
rawx: Implement a buffer cache to avoid GC complications

jfsmig:
##### SUMMARY
A shared cache for large buffers is used for stream operations (upon uploads)
Another shared cache for small buffers is used for xattr operations.

##### ISSUE TYPE
enhancement

##### COMPONENT NAME
rawx

##### SDS VERSION
something like 5.3

### DIFF
--- a/rawx/handler_chunk.go
+++ b/rawx/handler_chunk.go
@@ -100,7 +100,6 @@ func copyReadWriteBuffer(dst io.Writer, src io.Reader, pool bufferPool) (written
 
 	for {
 		// Fill the buffer
-		buf = buf[:cap(buf)]
 		totalr, er := fillBuffer(src, buf)
 
 		// Dump the buffer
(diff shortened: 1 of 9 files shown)


rawx/main.go
```diff
@@ -37,6 +37,8 @@ import (
 	"time"
 )
 
+var xattrBufferPool = newBufferPool(128*1024, 2048)
```
AymericDu, line 40: We should add these values in `const.go`.

rawx/handler_chunk.go
```diff
@@ -94,9 +94,13 @@ func dumpBuffer(dst io.Writer, buf []byte) (written int, err error) {
...
 }
 
-func copyReadWriteBuffer(dst io.Writer, src io.Reader, buf []byte) (written int64, err error) {
+func copyReadWriteBuffer(dst io.Writer, src io.Reader, pool bufferPool) (written int64, err error) {
+	buf := pool.Acquire()
+	defer pool.Release(buf)
+
 	for {
 		// Fill the buffer
+		buf = buf[:cap(buf)]
```
AymericDu, line 103: It's not already done with `Acquire`?

jfsmig: Yes it is! This is strain, a remainder of the development process.

PR #2031, opened 2020-05-07
conscience-agent: add timeout for http stats

murlock:
##### SUMMARY

It appears that some services may be seen as down by the conscience but
still running: the http stats plugin did not implement timeout on requests.

##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT NAME
conscience-agent

##### SDS VERSION
```
5.x
```

### DIFF
--- a/oio/conscience/stats/http.py
+++ b/oio/conscience/stats/http.py
@@ -1,4 +1,4 @@
-# Copyright (C) 2015-2018 OpenIO SAS, as part of OpenIO SDS
+# Copyright (C) 2015-2020 OpenIO SAS, as part of OpenIO SDS
 #
 # This program is free software: you can redistribute it and/or modify
 # it under the terms of the GNU Affero General Public License as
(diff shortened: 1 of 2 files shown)


oio/conscience/stats/http.py
```diff
@@ -75,9 +78,13 @@ def get_stats(self):
...
-            resp = self.agent.pool_manager.request(
-                'GET', self.url, headers={'Connection': 'close',
-                                          REQID_HEADER: reqid})
+            try:
+                with Timeout(self.timeout):
+                    resp = self.agent.pool_manager.request(
+                        'GET', self.url, headers={'Connection': 'close',
+                                                  REQID_HEADER: reqid})
+            except Timeout as e:
+                raise Exception(str(e))
```
fvennetier, line 87: If one stat fetcher raises an exception, the next stat fetchers won't run. There is something to fix in `oio/conscience/agent.py`.

PR #2033, opened 2020-05-07
cli: expose perfdata as dedicated entry

murlock:
##### SUMMARY

To allow use of `--dump-perfdata` from other tools,
the perdata will be exposed as dedicated columns in output.
The previous output is still available through the `--debug` flag.
You should use format yaml or json for best output

Fixes #OS-520

##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT NAME
cli

##### SDS VERSION
```
5.x
```

##### ADDITIONAL INFORMATION

Best display with yaml formatter
```
openio --dump-perfdata object create a /etc/magic -f yaml
- Hash: 272913026300E7AE9B5E2D51F138E674
  Name: magic
  Perfdata:
    data_size: 111
    proxy:
      meta2: 0.002684
      overall: 0.0062420000322163105
      resolve: 0.001771
    rawx:
      http://127.0.0.1:6010/92C08C2F08D20542F462965E0600154179761EA4B2F83B763F9345A4FAC68B14: 0.0006470000371336937
      overall: 0.0011420000810176134
  Size: 111
  Status: Ok
```

### DIFF
--- a/oio/cli/object/object.py
+++ b/oio/cli/object/object.py
@@ -169,6 +169,7 @@ class CreateObject(ContainerCommandMixin, Lister):
         any_error = False
         properties = parsed_args.property
         results = []
+        perfdata = self.app.client_manager.storage.perfdata
         for obj in objs:
             name = obj
             try:
@@ -187,7 +188,11 @@ class CreateObject(ContainerCommandMixin, Lister):
                         mime_type=parsed_args.mime_type,
                         autocreate=autocreate)
 
-                    results.append((name, data[1], data[2].upper(), 'Ok'))
+                    res = (name, data[1], data[2].upper(), 'Ok')
+                    if perfdata:
+                        res += (perfdata,)
+
+                    results.append(res)
             except KeyboardInterrupt:
                 results.append((name, 0, None, 'Interrupted'))
                 any_error = True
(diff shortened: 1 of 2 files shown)


oio/cli/object/object.py
```diff
@@ -187,7 +188,11 @@ def take_action(self, parsed_args):
                         mime_type=parsed_args.mime_type,
                         autocreate=autocreate)
 
-                    results.append((name, data[1], data[2].upper(), 'Ok'))
+                    res = [name, data[1], data[2].upper(), 'Ok']
+                    if perfdata:
+                        res += (perfdata,)
```
fvennetier, line 193: No need to wrap this into a tuple: the element will be added to the list anyway.
Or we can make both elements tuples:
```
res = (name, data[1], data[2].upper(), 'Ok')
if perfdata:
    res = res + (perfdata, )
```

oio/cli/object/object.py
```diff
@@ -199,7 +204,9 @@ def take_action(self, parsed_args):
                 results.append((name, 0, None, 'Failed'))
 
         listing = (obj for obj in results)
-        columns = ('Name', 'Size', 'Hash', 'Status')
+        columns = ['Name', 'Size', 'Hash', 'Status']
+        if perfdata:
+            columns += ('Perfdata',)
```
fvennetier, line 209: Same remark here.

PR #2045, opened 2020-05-28
rawx: add more logs when replying error

murlock:
##### SUMMARY

When rawx returns an error, there is only few codepath where more info is
logged as error.

I've chosen to always log error when we force connection to be closed instead of
adding LogError before each rr.replyWithError. 

##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT NAME
rawx

##### SDS VERSION
```
7.0.0.0b2-dev1
```

##### ADDITIONAL INFORMATION

```

```

### DIFF
--- a/rawx/handler_chunk.go
+++ b/rawx/handler_chunk.go
@@ -154,7 +154,7 @@ func (rr *rawxRequest) uploadChunk() {
 	var h hash.Hash
 
 	if rr.chunk, err = retrieveHeaders(&rr.req.Header, rr.chunkID); err != nil {
-		rr.replyError(err)
+		rr.replyError("uploadChunk()", err)
 		// Discard request body
 		io.Copy(ioutil.Discard, rr.req.Body)
 		return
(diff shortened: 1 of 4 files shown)


rawx/handler_chunk.go
```diff
@@ -245,25 +245,25 @@ func (rr *rawxRequest) uploadChunk() {
...
-		rr.replyError(err)
+		rr.replyError("copyChunk()", err)
 	} else {
 		// Link created, try to place an xattr
 		err = rr.chunk.saveContentFullpathAttr(op)
 		if err != nil {
 			// Xattr failed, rollback the link itself
 			LogError(msgErrorAction("Setxattr()", rr.reqid, err))
-			rr.replyError(err)
+			rr.replyError("copyChunk()", err)
```
fvennetier, line 266: Won't we report the same error message twice?

murlock: Good catch !

PR #2051, opened 2020-06-04
Check size of all uploaded chunks

AymericDu:
##### SUMMARY

Check size of all uploaded chunks

##### ISSUE TYPE

 - Bugfix Pull Request

##### COMPONENT NAME

- Python API
- `blob-mover`
- `blob-rebuilder`

##### SDS VERSION

```
openio 5.5.3.dev5
```

### DIFF
--- a/tests/functional/blob/test_mover.py
+++ b/tests/functional/blob/test_mover.py
@@ -14,12 +14,14 @@
 # License along with this library.
 
 import random
+from mock import MagicMock as Mock
 
 from oio.api.object_storage import ObjectStorageApi
 from oio.blob.client import BlobClient
-from oio.common.utils import cid_from_name
+from oio.common.utils import GeneratorIO, cid_from_name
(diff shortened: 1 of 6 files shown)


tests/functional/blob/test_mover.py
```diff
@@ -118,3 +121,25 @@ def test_move_old_chunk(self):
...
+        orig_chunk = random.choice(self.chunks)
+        chunk_volume = orig_chunk['url'].split('/')[2]
+        chunk_id = orig_chunk['url'].split('/')[3]
+
+        mover = BlobMoverWorker(self.conf, None,
+                                self.rawx_volumes[chunk_volume])
+        meta, stream = mover.blob_client.chunk_get(orig_chunk['url'])
+        data = stream.read()
+        stream.close()
+        data = data[:-1]
```
fvennetier, line 138: If someday you have to do that in a big stream, you can use `itertools.islice()`.

PR #2055, opened 2020-06-15
xcute: fix job listing in python3

loamhoof:
##### SUMMARY
Fix xcute job listing not working in Python3.

##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT NAME
xcute

##### ADDITIONAL INFORMATION
The problem came from stringifying bytes, which become `b'my_string'` when converting into a string, instead of the wanted `my_string`.

### DIFF
--- a/oio/xcute/common/backend.py
+++ b/oio/xcute/common/backend.py
@@ -537,7 +537,7 @@ class XcuteBackend(RedisConnection):
 
             pipeline = self.conn.pipeline()
             for job_id in job_ids:
-                self._get_job_info(job_id, client=pipeline)
+                self._get_job_info(job_id.decode('utf-8'), client=pipeline)
             job_infos = pipeline.execute()
 
             for job_info in job_infos:


oio/xcute/common/backend.py
```diff
@@ -537,7 +537,7 @@ def list_jobs(self, marker=None, limit=1000):
 
             pipeline = self.conn.pipeline()
             for job_id in job_ids:
-                self._get_job_info(job_id, client=pipeline)
+                self._get_job_info(job_id.decode(), client=pipeline)
```
AymericDu, line 540: For clarity, the `encoding` parameter should be specified. Like here : https://github.com/open-io/oio-sds/blob/afefff8689b1e1fed3f880dbe5904b0131b22e41/oio/xcute/common/backend.py#L704

loamhoof: :+1:

PR #2057, opened 2020-06-19
Python API: Refresh the cached rawx scores

AymericDu:
##### SUMMARY

Refresh the rawx scores from cache

Refresh every 30 seconds (default) so as not to request the proxy all the time.

##### ISSUE TYPE

 - Bugfix Pull Request

##### COMPONENT NAME

- Python API

##### SDS VERSION

```
openio 5.5.3.dev7
```

### DIFF
--- a/oio/common/cache.py
+++ b/oio/common/cache.py
@@ -126,7 +126,14 @@ def set_cached_object_metadata(content_meta, content_chunks,
     content_meta['properties'] = dict()
     cache_value['meta'] = content_meta
     if content_chunks is not None:
-        cache_value['chunks'] = content_chunks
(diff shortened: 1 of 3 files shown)


oio/common/cache.py
```diff
@@ -126,7 +126,14 @@ def set_cached_object_metadata(content_meta, content_chunks,
     content_meta['properties'] = dict()
     cache_value['meta'] = content_meta
     if content_chunks is not None:
-        cache_value['chunks'] = content_chunks
+        downsized_chunks = list()
```
fvennetier, line 129: What?

PR #2062, opened 2020-07-03
Enhance bucket listing with versioning status

murlock:
##### SUMMARY

Add option to show versioning status for `bucket list` command.

##### ISSUE TYPE
 - Feature Pull Request

##### COMPONENT NAME
CLI

##### SDS VERSION
```
5.5.5.4.dev1
```

### DIFF
--- a/oio/cli/container/container.py
+++ b/oio/cli/container/container.py
@@ -18,6 +18,7 @@
 from logging import getLogger
 
 from oio.cli import Command, Lister, ShowOne
+from oio.common.exceptions import NoSuchContainer
 from oio.common.timestamp import Timestamp
 from oio.common.utils import depaginate, request_id, timeout_to_deadline
 from oio.common.constants import \
@@ -487,6 +488,12 @@ class ListBuckets(Lister):
                   "(and set output format to 'value')"),
             action=ValueFormatStoreTrueAction
         )
+        parser.add_argument(
+            '--versioning',
+            action='store_true',
+            dest='versioning',
+            help="Display the versioning state of each bucket"
+        )
         return parser
 
     def take_action(self, parsed_args):
@@ -502,6 +509,7 @@ class ListBuckets(Lister):
 
         account = self.app.client_manager.account
         acct_client = self.app.client_manager.storage.account
+        storage = self.app.client_manager.storage
 
         if parsed_args.full_listing:
             listing = depaginate(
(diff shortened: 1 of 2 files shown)


oio/cli/container/container.py
```diff
@@ -516,8 +525,28 @@ def take_action(self, parsed_args):
...
+            for v in listing:
+                try:
+                    data = storage.container_get_properties(account, v['name'])
+                except NoSuchContainer:
+                    v['versioning'] = "Error"
+                    self.log.info('Bucket %s does not exist', v['name'])
+                    continue
+
+                sys = data['system']
+                # TODO(mbo) we should use default namespace configuration
```
fvennetier, line 542: Enabling versioning by default on the whole namespace goes against S3 compatibility. This is a feature we should drop.

oio/cli/container/container.py
```diff
@@ -516,8 +525,28 @@ def take_action(self, parsed_args):
             listing = acct_meta['listing']
 
         columns = ('Name', 'Bytes', 'Objects', 'Mtime')
-        return columns, ((v['name'], v['bytes'], v['objects'], v['mtime'])
-                         for v in listing)
+
+        if parsed_args.versioning:
+            columns += ('Versioning',)
+            # TODO(mbo) rework to use benefits of yield if no-paging is enabled
```
fvennetier, line 531: Since there is no pressure on this new feature, I suggest we implement this before merging this pull request.

oio/cli/container/container.py
```diff
@@ -487,6 +488,13 @@ def get_parser(self, prog_name):
                   "(and set output format to 'value')"),
             action=ValueFormatStoreTrueAction
         )
+        parser.add_argument(
+            '--versioning',
+            action='store_true',
+            dest='versioning',
+            help="""Show status of versioning per bucket
+            like aws s3api get-bucket-versioning"""
```
fvennetier, line 496: "Display the versioning state of each bucket".

tests/functional/cli/container/test_container.py
```diff
@@ -84,6 +84,23 @@ def test_bucket_list(self):
...
+                             (cname, cname, opts))
+        self.assertOutput(cname + '\n', output)
+
+        opts = self.get_format_opts(fields=('Name', 'Versioning'))
+        opts += " --prefix %s --versioning" % cname
+
+        output = self.openio('bucket list ' + opts)
+        self.assertIn("Suspended", output)
+
+        self.openio('container set --max-versions -1 %s' % cname)
```
fvennetier, line 100: To be coherent, we can use `--versioning` instead of `--max-versions`.

murlock: Done

PR #2069, opened 2020-07-21
Properly pass the connection and write timeout when linking

AymericDu:
##### SUMMARY

Properly pass the connection and write timeout when linking

##### ISSUE TYPE

 - Bugfix Pull Request

##### COMPONENT NAME

- Python API

##### SDS VERSION

```
openio 5.5.5.dev3
```

### DIFF
--- a/oio/blob/client.py
+++ b/oio/blob/client.py
@@ -1,4 +1,4 @@
-# Copyright (C) 2015-2019 OpenIO SAS, as part of OpenIO SDS
+# Copyright (C) 2015-2020 OpenIO SAS, as part of OpenIO SDS
 #
 # This library is free software; you can redistribute it and/or
 # modify it under the terms of the GNU Lesser General Public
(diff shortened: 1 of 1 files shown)


oio/blob/client.py
```diff
@@ -133,19 +135,19 @@ def chunk_delete_many(self, chunks, cid=None,
             or `urllib3.exceptions.HTTPError`, with an extra "chunk"
             attribute.
         """
-        headers = kwargs['headers'].copy()
+        headers = kwargs.pop('headers', None).copy()
```
fvennetier, line 138: `None.copy()` will fail badly. No need to make a copy here.

oio/blob/client.py
```diff
@@ -190,18 +192,18 @@ def chunk_head(self, url, **kwargs):
         :returns: a `dict` with chunk metadata (empty when xattr is False).
         """
         _xattr = bool(kwargs.get('xattr', True))
-        url = self.resolve_url(url)
-        headers = kwargs['headers'].copy()
+        headers = kwargs.pop('headers', None).copy()
```
fvennetier, line 195: Same comment: no need to make a copy.

PR #2070, opened 2020-07-23
account: Refresh account without impacting the bucket counters

AymericDu:
##### SUMMARY

Refresh account without impacting the bucket counters
- Use the bucket name already registered when it is not given
- Add all totals when a container is not yet associated with its bucket
- Refresh bucket counters

Jira: OS-562

##### ISSUE TYPE

 - Bugfix Pull Request

##### COMPONENT NAME

- account

##### SDS VERSION

```
openio 5.5.5.dev7
```


murlock:
See with you, we should allow to set bucket-name property with already existing containers

fvennetier:
> See with you, we should allow to set bucket-name property with already existing containers

See #2073.

PR #2074, opened 2020-08-26
account: do not try to decode "None" as UTF-8

AymericDu:
##### SUMMARY
Do not try to decode "None" as UTF-8. This error was frequent but unfortunately it was ignored by our tests.

##### ISSUE TYPE

 - Bugfix Pull Request

##### COMPONENT NAME

- account

##### SDS VERSION

```
openio 7.0.2.dev13
```

### DIFF
--- a/oio/account/backend.py
+++ b/oio/account/backend.py
@@ -581,7 +581,8 @@ class AccountBackend(RedisConnection):
                 pass
         for what in (BUCKET_PROP_REPLI_ENABLED.encode('utf-8'), ):
             try:
-                decoded = info.get(what, b'').decode('utf-8')
+                val = info.get(what)
+                decoded = val.decode('utf-8') if val is not None else None
                 info[what] = boolean_value(decoded)
             except (TypeError, ValueError):
                 pass


oio/account/backend.py
```diff
@@ -581,7 +581,7 @@ def cast_fields(self, info):
                 pass
         for what in (BUCKET_PROP_REPLI_ENABLED.encode('utf-8'), ):
             try:
-                decoded = info.get(what, b'').decode('utf-8')
+                decoded = (info.get(what) or b'').decode('utf-8')
                 info[what] = boolean_value(decoded)
```
fvennetier, line 585: I'm not sure this fixes anything.
```
>>> from oio.common.easy_value import boolean_value
>>> boolean_value(b''.decode('utf-8'))
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/fvenneti/src/public_git/oio-sds/oio/common/easy_value.py", line 60, in boolean_value
    raise ValueError('Boolean value expected')
ValueError: Boolean value expected
```

```suggestion
                decoded = (info.get(what) or b'').decode('utf-8') or None
                info[what] = boolean_value(decoded)
```
Why `None`? Because it is the first value considered "false" by the `boolean_value` method.

fvennetier (approved):

PR #2080, opened 2020-09-09
Remove unsupported backblaze integration

murlock:
##### SUMMARY

Remove backblaze support as it was not tested since few years

##### ISSUE TYPE
 - (un)Feature Pull Request

##### COMPONENT NAME
- all

##### SDS VERSION
```
master
```

##### ADDITIONAL INFORMATION

```

```


meta2v2/meta2_utils_lb.c
```diff
@@ -192,9 +192,8 @@ static gchar*
 m2v2_build_chunk_url_storage (const struct storage_policy_s *pol,
 		const gchar *str_id)
 {
+	(void)(str_id);
 	switch(data_security_get_type(storage_policy_get_data_security(pol))) {
-	case STGPOL_DS_BACKBLAZE:
-		return g_strconcat("b2/", str_id, NULL);
```
fvennetier, line 197: Since there is only one choice, no need for a switch.

murlock: :+1:

fvennetier: By the way, the calling function will crash if this function returns NULL. I suggest removing this one, and modifying `generate_chunk_bean`.

PR #2085, opened 2020-09-29
account: rework bucket refresh

murlock:
##### SUMMARY

The compute of refresh for a bucket is now done by batch:
 - a lock key is created and will hold ctime, mtime and marker.
 - bucket info is set to 0
 - a batch will parses at most 10000 containers and will update
   stats.
 - when a container for this bucket is updated, it will check
  if lock key exists. If it exists, the bucket stats will be updated if container name is already parsed by refresh operation
  (name is before or equal to marker)

##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT NAME
account

##### SDS VERSION
```
5.6.1dev2
```

### DIFF
--- a/oio/account/backend.py
+++ b/oio/account/backend.py
@@ -19,6 +19,7 @@ import re
 import redis
 import redis.sentinel
 import six
+
 from werkzeug.exceptions import NotFound, Conflict, BadRequest
 from oio.common.constants import BUCKET_PROP_REPLI_ENABLED
 from oio.common.timestamp import Timestamp
@@ -30,6 +31,7 @@ ACCOUNT_KEY_PREFIX = 'account:'
 BUCKET_KEY_PREFIX = 'bucket:'
 BUCKET_LIST_PREFIX = 'buckets:'
 CONTAINER_LIST_PREFIX = 'containers:'
+BUCKET_LOCK_KEY_PREFIX = 'bucketlock:'
 
 EXPIRE_TIME = 60  # seconds
 
@@ -59,7 +61,7 @@ class AccountBackend(RedisConnection):
         -- (it has a prefix).
         local update_bucket_stats = function(
             container_key, bucket_key, buckets_list_key,
-            account, container_name, bucket_name, mtime, deleted,
+            account, container_name, bucket_name, bucket_lock, mtime, deleted,
             inc_objects, inc_bytes, inc_damaged_objects, inc_missing_chunks)
           if deleted then
             redis.call('HDEL', container_key, 'bucket');
@@ -90,13 +92,18 @@ class AccountBackend(RedisConnection):
             inc_objects = 0;
           end;
 
-          -- Increment the counters.
-          redis.call('HINCRBY', bucket_key, 'objects', inc_objects);
-          redis.call('HINCRBY', bucket_key, 'bytes', inc_bytes);
-          redis.call('HINCRBY', bucket_key, 'damaged_objects',
-                     inc_damaged_objects);
-          redis.call('HINCRBY', bucket_key, 'missing_chunks',
-                     inc_missing_chunks);
+          -- Check if a refresh bucket is in progress
+          local marker = redis.call("HGET", bucket_lock, "marker");
+
+          -- Increment the counters if needed.
+          if marker == false or container_name <= marker then
+            redis.call('HINCRBY', bucket_key, 'objects', inc_objects);
+            redis.call('HINCRBY', bucket_key, 'bytes', inc_bytes);
+            redis.call('HINCRBY', bucket_key, 'damaged_objects',
+                       inc_damaged_objects);
+            redis.call('HINCRBY', bucket_key, 'missing_chunks',
+                       inc_missing_chunks);
+          end;
 
           -- Finally update the modification time.
           if mtime ~= '' then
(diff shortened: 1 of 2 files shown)


oio/account/backend.py
```diff
@@ -918,9 +921,16 @@ def refresh_bucket(self, bucket_name, **kwargs):
                 self.clistkey(account_id), self.bkey(bucket_name),
                 self.blistkey(account_id)]
         args = [bucket_name]
+
         try:
-            self.script_refresh_bucket(
-                keys=keys, args=args)
+            values = self.script_refresh_bucket(keys=keys, args=args,
+                                                client=self.conn_slave)
+            self.conn.hmset(self.bkey(bucket_name), {
```
fvennetier, line 928: The purpose of the "bucket refresh" operation is to fix incoherences in the bucket stats. But this 2-step operation does not guarantee that the count will be coherent after its execution.

oio/account/backend.py
```diff
@@ -913,14 +980,36 @@ def refresh_bucket(self, bucket_name, **kwargs):
...
+                        self.clistkey(account_id), self.bkey(bucket_name),
+                        marker, lkey]
+                res = self.script_refresh_bucket(keys=keys, args=args)
+                if not res[0]:
+                    break
+                marker = res[0]
+                # could be done in LUA refresh
+                self.conn.hset(lkey,
+                               mapping={"marker": marker,
+                                        "mtime": int(time())})
```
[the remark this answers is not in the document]
AymericDu: As the marker is already set in the `refresh_bucket` script, we should give the `mtime` directly as arguments to this script.

oio/account/backend.py
```diff
@@ -913,14 +980,36 @@ def refresh_bucket(self, bucket_name, **kwargs):
...
+            self.conn.delete(lkey)
```
AymericDu, line 1012: The lock should be released directly in the `refresh_bucket` script so as not to forget to increment new containers that are greater than the last marker.
Therefore, we should also take the lock directly from the `refresh_bucket` script.

oio/account/backend.py
```diff
@@ -147,20 +170,33 @@ class AccountBackend(RedisConnection):
...
+        -- Increment the counters.
+        redis.call('HINCRBY', bkey, 'objects', total_objects);
+        redis.call('HINCRBY', bkey, 'bytes', total_bytes);
+        redis.call('HINCRBY', bkey, 'damaged_objects',
+                    total_damaged_objects);
+        redis.call('HINCRBY', bkey, 'missing_chunks',
+                    total_missing_chunks);
+
+        redis.call('HSET', lkey, 'marker', new_marker);
+        return { new_marker };
```
AymericDu, line 197: We could only return a boolean: `return { i ~= batch_size }`

oio/account/backend.py
```diff
@@ -121,17 +129,32 @@ class AccountBackend(RedisConnection):
...
-                   'bytes', 0,
-                   'damaged_objects', 0,
-                   'missing_chunks', 0);
+        -- global counters
+        local total_objects = 0;
+        local total_bytes = 0;
+        local total_damaged_objects = 0;
+        local total_missing_chunks = 0;
+
+        local batch_size = 10000;
```
AymericDu, line 139: We could make it a parameter.

oio/account/backend.py
```diff
@@ -147,20 +170,33 @@ class AccountBackend(RedisConnection):
...
         end;
-        """)
+
+        -- Increment the counters.
+        redis.call('HINCRBY', bkey, 'objects', total_objects);
+        redis.call('HINCRBY', bkey, 'bytes', total_bytes);
+        redis.call('HINCRBY', bkey, 'damaged_objects',
+                    total_damaged_objects);
+        redis.call('HINCRBY', bkey, 'missing_chunks',
+                    total_missing_chunks);
```
AymericDu, line 194: We should also update the `mtime`.
```suggestion
                    total_missing_chunks);

        -- Finally update the modification time.
        redis.call('HSET', bkey, 'mtime', mtime);
```

oio/account/backend.py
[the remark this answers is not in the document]
murlock: :+1:

oio/account/backend.py
murlock: I don't get your point but I'll move `self.conn.delete(lkey)` in a `finally:` block to be sure that lock is properly removed.

oio/account/backend.py
murlock: The `mtime` of Bucket is last modification for object operations, I don't think that "refresh bucket" is a standard operation that should update it. Let's talk if we are not agree ^^

oio/account/backend.py
murlock: :+1:

oio/account/backend.py
murlock: yes, we could ^^

oio/account/backend.py
AymericDu: It isn't false.

oio/account/backend.py
AymericDu: With an example, it will be clearer:
- the `refresh_bucket` script ended with the last `toto` marker
- the transaction is complete, so new requests can be executed on the master
- a new `tyty` container is created
- `"tyty" > "toto"`, so the bucket stats are not updated
- the lock key is deleted
- the bucket stats are not correct

Another solution can be to set a fake marker (`"\xef\xbf\xbd"`) at the end of the script when the refresh is finished.

murlock: Gotcha !

AymericDu (approved):

PR #2089, opened 2020-10-15
[WIP] Implement container sharding (first step)

AymericDu:
##### SUMMARY

- [x] Add shard container in root container
- [x] Redirect an object request to the shard container
- [x] Shard empty container
- [x] PUT/GET an object in a sharded container
- [x] List the shards container of root container
- [x] List the objects in a sharded container
- [ ] Make sharding compatible with versioning

##### ISSUE TYPE

 - Feature Pull Request

##### COMPONENT NAME

- `meta2`
- `proxy`
- CLI

##### SDS VERSION

```
openio 8.0.0.dev1
```

##### Example

```
$ openio container create mycontainer
+------+---------+
| Name | Created |
+------+---------+
| toto | True    |
+------+---------+
$ openio container-sharding replace mycontainer '[{"index": 0, "lower": "", "upper": "sharding"}, {"index": 1, "lower": "sharding", "upper": ""}]' --enable
+-------+----------+----------+
| Index | Lower    | Upper    |
+-------+----------+----------+
|     0 |          | sharding |
|     1 | sharding |          |
+-------+----------+----------+
$ openio container-sharding show mycontainer --counts
+-------+----------+----------+------------------------------------------------------------------+-------+-------+
| Index | Lower    | Upper    | CID                                                              | Bytes | Count |
+-------+----------+----------+------------------------------------------------------------------+-------+-------+
|     0 |          | sharding | B1348A230C6B88ED18232EA0A877B2455732ECD158E01EE655289A07E6C55D2F | 0     | 0     |
|     1 | sharding |          | 79110CB5BDF6327D4FD344828DC3E724E615D9855C885B447CB5ED64B15ADF20 | 0     | 0     |
+-------+----------+----------+------------------------------------------------------------------+-------+-------+
$ openio object create mycontainer /etc/magic --name magic
+-------+------+----------------------------------+--------+
| Name  | Size | Hash                             | Status |
+-------+------+----------------------------------+--------+
| magic |  111 | 272913026300E7AE9B5E2D51F138E674 | Ok     |
+-------+------+----------------------------------+--------+
$ openio container-sharding show mycontainer --counts
+-------+----------+----------+------------------------------------------------------------------+-------+-------+
| Index | Lower    | Upper    | CID                                                              | Bytes | Count |
+-------+----------+----------+------------------------------------------------------------------+-------+-------+
|     0 |          | sharding | B1348A230C6B88ED18232EA0A877B2455732ECD158E01EE655289A07E6C55D2F | 111   | 1     |
|     1 | sharding |          | 79110CB5BDF6327D4FD344828DC3E724E615D9855C885B447CB5ED64B15ADF20 | 0     | 0     |
+-------+----------+----------+------------------------------------------------------------------+-------+-------+
$ openio object create mycontainer /etc/magic --name test
+------+------+----------------------------------+--------+
| Name | Size | Hash                             | Status |
+------+------+----------------------------------+--------+
| test |  111 | 272913026300E7AE9B5E2D51F138E674 | Ok     |
+------+------+----------------------------------+--------+
$ openio container-sharding show mycontainer --counts
+-------+----------+----------+------------------------------------------------------------------+-------+-------+
| Index | Lower    | Upper    | CID                                                              | Bytes | Count |
+-------+----------+----------+------------------------------------------------------------------+-------+-------+
|     0 |          | sharding | B1348A230C6B88ED18232EA0A877B2455732ECD158E01EE655289A07E6C55D2F | 111   | 1     |
|     1 | sharding |          | 79110CB5BDF6327D4FD344828DC3E724E615D9855C885B447CB5ED64B15ADF20 | 111   | 1     |
+-------+----------+----------+------------------------------------------------------------------+-------+-------+
```


oio/container/sharding.py
```diff
@@ -0,0 +1,213 @@
...
+            raise ValueError('Expected a string for the "upper"')
+        new_shard['upper'] = shard_upper
+
+        if shard_lower != '' and shard_upper != '' \
+                and shard_lower >= shard_upper:
+            raise ValueError('Expected a "upper" greater the "lower"')
+
+        if timestamp:
+            shard_timestamp = shard.get('timestamp')
+            if shard_upper is None:
```
murlock, line 91: should be `shard_timestamp` IMO ?

oio/container/sharding.py
```diff
@@ -0,0 +1,213 @@
...
+                     cid=None, **kwargs):
+        if cid:
+            params = {'cid': cid}
+        else:
+            params = {'acct': account, 'ref': reference}
+        if path:
+            params.update({'path': path})
+        return params
+
+    def _format_shard(self, shard, timestamp=False):
```
murlock, line 46: I don't see how timestamp is used here.
A little explanation is welcome :)

oio/container/sharding.py
```diff
@@ -0,0 +1,213 @@
...
+        for shard in shards:
+            new_shards.append(self._format_shard(shard))
+
+        new_shards.sort(key=lambda new_shard: new_shard['index'])
+        first_shard_index = new_shards[0]['index']
+        last_shard_index = new_shards[-1]['index']
+        previous_shard_upper = ''
+        for i, new_shard in enumerate(new_shards):
+            if new_shard['index'] != i:
+                raise ValueError('Missing "index" %d' % i)
```
murlock, line 121: IMO error message is wrong, it should be something like "Invalid `"index" %d, found %d" % (new_shard['index'], i)`

oio/cli/containersharding/sharding.py
```diff
@@ -0,0 +1,85 @@
...
+        parser = super(ReplaceContainerSharding, self).get_parser(prog_name)
+        self.patch_parser_container_sharding(parser)
+        parser.add_argument(
+            'configuration',
+            metavar='<configuration>',
+            help="""
+                Sharding configuration
+                JSON Syntax:
+                [{"index": 0, "lower": "", "upper": "sharding"},
+                 {"index": 1, "lower": "sharding", "upper": ""}]
```
murlock, line 50: Since `upper` of Nth shard must be the `lower` of Nth+1 shard,
we may ask to user to provide
```
{"index": "0", "upper": "cde"},
{"index": "1, "upper": "ghi"},
{"index": "2", "upper": "opq"},
{"index": "3:", "upper": ""},
```
and build the full JSON internally, 
It will simplify various checks in ContainerSharding

murlock (commented):
Few remarks

oio/cli/containersharding/sharding.py
AymericDu: Unless the user wishes, this JSON will be created by `openio container-sharding find` (not yet implemented), as for `swift`.
In addition, having the `lower` and `upper` ensures that there has been no input error.

oio/container/sharding.py
AymericDu: The shards are sorted by index. So if the index is different than expected, this index is missing.

murlock (commented):
Some remarks:
- `openio container list` should show if a container is using sharding mecanism or not
- `openio container-sharding show --json` is not compatible with `openio container-sharding replace --from-file`
- ` openio container show` shoud show if sharding mecanism is used or not
- `openio container delete` should have a recurse / force flag and must fails if at least one shard is not empty ! (**it is possible to remove a bucket with objects inside !**)
- the shard contains timestamp as float `1608627740.169801`, IMO it will better to use integer

In case of errors, the message should be improved:
I've enabled sharding on a container and removed manually container shard
- when I put an object to this shard, error message still reference root container.
- the listing of root container is totally broken (I can't see the listing of first shard) and error message still reference root container.
- when replacing schema shard  by another one, the CLI doesn't show any error. Only if I set `-vv --debug`, I can see the expected `NotImplementedError`

AymericDu:
The pull request has been transferred to Stash.

PR #2093, opened 2020-11-12
multi-conscience: Keep the service locks when restarting conscience

AymericDu:
##### SUMMARY

Keep the service locks when restarting conscience with multi-conscience.

##### ISSUE TYPE

 - Bugfix Pull Request

##### COMPONENT NAME

- multi-conscience

##### SDS VERSION

```
openio 5.6.2.dev1
```

### DIFF
--- a/cluster/module/server.c
+++ b/cluster/module/server.c
@@ -1494,15 +1494,6 @@ restart_srv_from_file(gchar *path)
 
 	if (!g_file_test(path, G_FILE_TEST_EXISTS)) {
 		return FALSE;
-	} else if (hub_group_size > 0) {
-		GRID_NOTICE("Conscience hub enabled "
-				"(%d other nodes), "
-				"ignoring persistence file %s, "
-				"waiting to receive a service list.",
-				hub_group_size, path);
-		for (int i = 8; i > 0 && !hub_working; i--)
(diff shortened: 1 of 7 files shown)


murlock (approved):
LGTM
maybe we could enrich `tests/func/test_cluster.c` with few unit tests ?

cluster/module/server.c
```diff
@@ -724,9 +748,158 @@ _alert_service_with_zeroed_score(struct conscience_srv_s *srv)
...
+			struct service_info_dated_s));
+	sid->si = g_malloc0(sizeof(struct service_info_s));
+	conscience_srv_fill_srvinfo(sid->si, srv);
+	sid->lock_mtime = srv->lock_mtime;
+	sid->tags_mtime = srv->tags_mtime;
+	return sid;
+}
+
+static struct conscience_srv_s *
+conscience_srvtype_refresh_from_conscience(
```
fvennetier, line 766: The "_from_conscience" in the name of these functions is confusing. I guess the difference between these functions and the other ones (without the "_from_conscience" suffix) is that they act on "dated" service descriptions, and will do date comparisons before overwriting known service descriptions.

fvennetier (approved):

PR #2094, opened 2020-11-18
rawx: Check repository with probe

AymericDu:
##### SUMMARY

Check rawx repository with probe (as for metaX)

##### ISSUE TYPE

 - Bugfix Pull Request

##### COMPONENT NAME

- `rawx` (Golang)

##### SDS VERSION

```
openio 5.7.1.dev2
```

### DIFF
--- a/rawx/main.go
+++ b/rawx/main.go
@@ -24,6 +24,7 @@ http handler.
 import (
 	"context"
 	"flag"
+	"fmt"
 	"log"
 	"net"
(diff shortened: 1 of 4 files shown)


murlock (approved):
It should be nice if at startup (after checking xattr), we store the `fsid` and in `taskProbeRepository`, check if `fsid` has changed

fvennetier (approved):

PR #2095, opened 2020-11-18
CI: convert remaining tests using python 2 to python 3.6

murlock:
##### SUMMARY

Drop Python 2
Since gateway will use Python 3.7, launch tests with Python 3.7

##### ISSUE TYPE

 - Feature Pull Request
 - Bugfix Pull Request
 - Docs Pull Request

##### COMPONENT NAME
oio

##### SDS VERSION
```
master
```

### DIFF
--- a/tools/benchmark/oio-bench-object-create.py
+++ b/tools/benchmark/oio-bench-object-create.py
@@ -99,7 +99,7 @@ def main(threads, delay=5.0, duration=60.0):
 
 def _object_upload(ul_handler, **kwargs):
     ul_chunks = ul_handler.chunk_prep()
-    return ul_chunks.next(), 0, 'd41d8cd98f00b204e9800998ecf8427e'
+    return next(ul_chunks), 0, 'd41d8cd98f00b204e9800998ecf8427e'
 
 
 USAGE = """Concurrently create many fake objects in the same container.
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,16 +2,15 @@ sudo: required
 dist: bionic
 language: python
 python:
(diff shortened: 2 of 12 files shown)


fvennetier:
Building with both Python 3.6 and 3.7 seems overkill. I suggest we reduce the number of test environments.

ilitzroth:
Please keep me informed about the python version. I'm currently using 3.6 if we're moving
to 3.7 some things need to be adapted (docker files, CMakeLists.txt), but I don't expect much fallout.

murlock:
We still target Python 3.6 at this time, I'll fix the error and reduce number of test environmment.

fvennetier:
Now that #2096 is merged, we need to make a cascade merge, and the tests should be ok again.

tools/benchmark/oio-bench-object-create.py
```diff
@@ -99,7 +99,7 @@ def main(threads, delay=5.0, duration=60.0):
 
 def _object_upload(ul_handler, **kwargs):
     ul_chunks = ul_handler.chunk_prep()
-    return ul_chunks.next(), 0, 'd41d8cd98f00b204e9800998ecf8427e'
+    return next(ul_chunks.next), 0, 'd41d8cd98f00b204e9800998ecf8427e'
```
AymericDu, line 102: I think there is one `next` too many.
```suggestion
    return next(ul_chunks), 0, 'd41d8cd98f00b204e9800998ecf8427e'
```

AymericDu (approved):

AymericDu (approved):

PR #2098, opened 2020-11-26
proxy: map more low-level errors to Service Unavailable (503)

fvennetier:
##### SUMMARY
Some low-level errors like CODE_AVOIDED (8) did not have a particular mapping to high-level errors, and thus triggered the ambiguous Internal Server Error (500) error. We choose to map these error to the Service Unavailable (503) which is better managed by calling applications (they usually retry the request after a delay).

##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT NAME
- oio-proxy

##### SDS VERSION
```
openio 5.7.1.dev1
```


murlock (approved):
LGTM

AymericDu (approved):

PR #2100, opened 2020-11-30
conscience-agent: check for service endpoint conflicts

fvennetier:
##### SUMMARY
If several services are configured on the same endpoint, log an error,
and do not report them to the conscience.

Jira: OBSTO-192

##### ISSUE TYPE
 - Feature Pull Request

##### COMPONENT NAME
- conscience-agent

##### SDS VERSION
```
openio 7.1.1.dev2
```

### DIFF
--- a/oio/conscience/agent.py
+++ b/oio/conscience/agent.py
@@ -214,6 +214,7 @@ class ConscienceAgent(Daemon):
         self.conf = conf
         self.logger = get_logger(conf)
         self.load_services()
+        self.check_for_conflicts()
         self.init_watchers(self.conf['services'])
 
     def stop(self):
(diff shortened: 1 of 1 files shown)


AymericDu (approved):

cdelgehier:
So first come, first served ? 
I didn't see any unregister action, so is it a "first come, first served" ?

fvennetier:
> So first come, first served ?
> I didn't see any unregister action, so is it a "first come, first served" ?

That's right, I did not put any unregister action. None of the conflicting services will be updated. But I can change that, and register one of them. Notice though that the code cannot identify which one is the good one...

fatpat:
would it be possible to lock them with a 0 score ?

fvennetier:
> would it be possible to lock them with a 0 score ?

It is possible. But the problem is that "locking" them means registering them. If a faulty service is "locked", it will appear in the conscience until it is garbage collected. With this pre-registration check, it won't be registered at all.

murlock (approved):

fatpat (approved):

